PR #1, opened 2021-09-28
Repository kickoff, with mtcli and lots of scaffolding.

sugarraysam:
Adding Makefile, Dockerfiles, app-interface integration.

The operator is a long-shot project.

Will be working mostly on `cmd/mtcli/main.go` to do index + bundle + imageset validation..

The goal is to import this binary in our tooling:

```python
# example: https://gitlab.cee.redhat.com/service/managed-tenants-bundles/-/blob/main/ci.py#L408
from sretoolbox.binaries import Mtcli # needs to be added to sretoolbox

MTCLI = Mtcli(version='x.x.x', download_path='/tmp')
MTCLI.run(*cmd)
```

app-interface integration MR (WIP):


api/v1alpha1/validation.go
```diff
@@ -0,0 +1,51 @@
...
+	// no errors
+	return nil
+}
+
+// PrintValidationErrors - helper to pretty print validationErrors and fail
+func PrintValidationErrors(fieldErrors *[]validator.FieldError) {
+	for _, err := range *fieldErrors {
+		fmt.Println(err)
+	}
+}
```
mayankshah1607, line 51: Does it make more sense to move this to a separate utils package?

api/v1alpha1/helpers.go
```diff
@@ -0,0 +1,20 @@
...
+import "k8s.io/apimachinery/pkg/util/yaml"
+
+// *****
+// Helper types
+// *****
+// Channel - list all channels for a given operator
+type Channel struct {
+	Name       string `json:"name"`
+	CurrentCSV string `json:"currentCSV"`
+}
```
mayankshah1607, line 12: Should this be in `addonmetadata_types.go` instead?

mayankshah1607 (commented):
Thanks for this, @sugarraysam ! 🚀 : Overall looks great to me, just had a few concerns which I've mentioned below :)

build_tag.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+set -exvo pipefail -o nounset
+
+# Simple script to download and run goreleaser
+# uses config from .goreleaser.yml
+curl -sL https://git.io/goreleaser | bash
```
mayankshah1607, line 7: I'm a bit confused as to how this will work.

How does it know which binaries to build? I don't see it mentioned in the `.goreleaser.yml`.

Does this also publish the binary against an already existing release, or does it create a new release for you? app-intercace's `gh-build-tag` is triggered only upon creating a new release tag.

api/v1alpha1/helpers.go
```diff
@@ -0,0 +1,20 @@
...
+
+// *****
+// Helper funcs/methods
+// *****
+// FromYAML - instantiates an AddonMetadata struct from yaml data
+func (a *AddonMetadata) FromYAML(data []byte) error {
+	return yaml.Unmarshal(data, a)
+}
```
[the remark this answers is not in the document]
sugarraysam: I often used this pattern of `FromXXX` as a method receiver on a type (e.g.: `FromYAML`, `FromJSON`, etc.). Because it is bound to a specific version `v1alpha` I don't see where else I could expose these. As declaring method receivers outside of a package is prohibited... :thinking:

api/v1alpha1/validation.go
sugarraysam: done

build_tag.sh
sugarraysam: I need to cleanup `.goreleaser.yml`. It is missing instructions on where to find the binary `cmd/mtcli/main.go` and probably other stuff. I've only copied a base file for now but this is indeed incomplete. 

You are right `gh-build-tag` is triggered after pushing a new tag. So whenever we want to trigger a release, we need to manually tag and push. You can see an example here: 
- https://github.com/app-sre/aws-resource-exporter/blob/master/build_tag.sh
- https://github.com/app-sre/aws-resource-exporter/tags

sugarraysam: Fixed and made sure goreleaser works as expected:
```bash
$ git tag v0.0.1
$ goreleaser check
$ goreleaser release --skip-publish
$ ./dist/mtcli_linux_amd64/mtcli version
```

api/v1alpha1/helpers.go
sugarraysam: moved

mayankshah1607:
> The goal is to import this binary in our tooling:

BTW, I'd be happy to help with integrating this binary into our python tooling once this PR is merged :)

mayankshah1607:
Next we'll have to validate things listed in https://issues.redhat.com/browse/MTSRE-209

I'm thinking we expose some kind of interface which each of the validations can implement (just brainstorming, I'm open to learning about better ideas :) )

I could help by adding code to validate a few things listed in the ticket

api/v1alpha1/helpers.go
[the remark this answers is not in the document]
mayankshah1607: Hmm, how about having a helper function like `GetAddonMetadataFromYAML(a *apis.AddonMetadata, data [byte]) error { ... }` ?

sugarraysam:
I've been keeping a list. Posted here as an issue: https://github.com/mt-sre/addon-metadata-operator/issues/2. Feel free to implement any of these !

As for the interface I PM'ed you !

pkg/objects/objects.go
```diff
@@ -0,0 +1,48 @@
...
+			},
+		},
+	}
+	return sss, nil
+}
+
+// NewDeleteSelectorSyncSet - creates a delete SelectorSyncSet from the addonMetadata
+func NewDeleteSelectorSyncSet(addonMetadata *addonsv1alpha1.AddonMetadata) (*hivev1.SelectorSyncSet, error) {
+	return nil, nil
+}
```
mayankshah1607, line 43: Are we calling these anywhere? I couldn't find any ref

api/v1alpha1/helpers.go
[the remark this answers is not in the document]
sugarraysam: If we bum to `v1beta1` wouldn't that break this function? seems it should live alongside each CRD version?

pkg/objects/objects.go
sugarraysam: This was just me messing around. I can remove it or comment it out. The idea was to replace our current jinja2 template with a strongly typed SelectorSyncSet.

api/v1alpha1/helpers.go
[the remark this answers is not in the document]
mayankshah1607: Yeah, sounds right. Let's just keep this one for now :)

pkg/objects/objects.go
sugarraysam: deleted

PR #4, opened 2021-10-19
Provide initial bootstrapping for validation CLI

mayankshah1607:
This PR adds some initial code to bootstrap the validations that will be run as a part of the CLI.

Screenshot 2021-10-20 at 4 01 01 PM

Signed-off-by: Mayank Shah <m.shah@redhat.com>

### DIFF
--- a/pkg/validators/validator_addon_label.go
+++ b/pkg/validators/validator_addon_label.go
@@ -13,7 +13,3 @@ func ValidateAddonLabel(metabundle *utils.MetaBundle) (bool, error) {
 
 	return true, nil
 }
-
-func someOtherUtilFunction() {
-
-}
--- a/pkg/validators/validate_something.go
+++ b/pkg/validators/validate_something.go
@@ -10,7 +10,3 @@ func ValidateSomethingCrossBetweenImageSetAndMetadata(metabundle *utils.MetaBund
 	}
 	return true, nil
 }
-
-func someUtilFunction() {
-
-}


sugarraysam (changes requested):
We had a design session with @yashvardhan-kukreja yesterday, and I agree with your comments we should probably drop go-playground/validator for now.

I really liked the end design, it had an interface and a common function signature for validators. Really simple and flexible. You guys should sync and review this code!

ashishmax31:
@yashvardhan-kukreja , I like the idea of just implementing validation functions which implement an interface, that way anyone can come-in and add new validation fns.

One thing we'd miss without the `go-playground/validator` is that, Our addon-metadata operator's CRDs would have marker based validations which can easily be copied to the struct's fields for the go-playground/validator to validate. (free validations w/o writing a single line of code).
Wdyt @mayankshah1607 @yashvardhan-kukreja @sugarraysam

sugarraysam:
Left a lengthy comment in JIRA ticket. Will not copy it here again, please have a look!

> One thing we'd miss without the go-playground/validator is that, Our addon-metadata operator's CRDs would have marker based validations which can easily be copied to the struct's fields for the go-playground/validator to validate. (free validations w/o writing a single line of code).

We could simply add a single function to do tag base go-playground style validation:

```go
type ValidationObjects struct {
  AddonMetadata *AddonMetadata `validate: "require"`
  Bundle *Bundle `validate: "require"`
  ImageSet *ImageSet `validate: "require"`
}


func(obj ValidationObjects) (error,string) {
  // single place where we use it - dont have to be global
  validator = validator.New()

  // recursive tag-level based validation (will dive in structs)
  err := validator.Struct(obj)

  // interpret err, get str etc.
  return err, "some message"
}
```

yashvardhan-kukreja:
> @yashvardhan-kukreja , I like the idea of just implementing validation functions which implement an interface, that way anyone can come-in and add new validation fns.
> 
> One thing we'd miss without the `go-playground/validator` is that, Our addon-metadata operator's CRDs would have marker based validations which can easily be copied to the struct's fields for the go-playground/validator to validate. (free validations w/o writing a single line of code).
> Wdyt @mayankshah1607 @yashvardhan-kukreja @sugarraysam 
> 

Nice catch Ashish but I thought about that as well and that won't be a concern.

We can still move ahead and tag the structs with go-validator's tags and that will give the new engineer/developer, while coding their validator, an opportunity to still make use of those tag-level validations by calling the `validator.Struct(<object>)` method inside the validator function which they shall be writing.

And if someone else, trying to add another validator, doesn't want to make use of go-validator and those tag-level validations, they can just simply choose to skip tag-level validation by not calling the go-validator's method (`validator.Struct(<object>)`) for tag-level validation. And hence, they can code out their validator in whatever manner they want to :)

yashvardhan-kukreja:
> I have some ideas and proposals around that and will be raising them as a PR by tomorrow on this repo :)

I just checked out the commit which was pushed by Mayank. And that commit is accomplishing the exact proposal (`ValidatorFunc`) which I was suggesting and wanted to implement. So, I am totally in favor with the latest state of this PR.

yashvardhan-kukreja:
Just checked out the latest commit by @mayankshah1607 . It's pretty much exactly in line with what I was suggesting (`ValidatorFunc`). So, thanks for implementing that out :)

I just had one suggestion here for the design of ValidatorFunc.
I suppose it would be better to make `ValidatorFunc` have this signature instead `ValidateFunc func(mb *MetaBundle) (bool, error)`

The bool denotes success/failure of validation (true -> success, false -> failure)
the error denotes any error which occurred while executing the ValidatorFunc.
Feels more semantic to me.

What do you folks think? :)

yashvardhan-kukreja:
> BTW, I've provided some boiler plate for color-coding logs and stuff too.
Just checked that out. Looks fab!

> Please feel free to fork this branch and push changes as you feel. We will get this merged once it looks good :)
Regarding I have already forked your forked repo instead. And I'll be going to make a PR against `mayankshah1607:mayank/implement-validation`
And if it looks good, you can approve and merge it in the future (with mayank/implement-validation branch), its commit/commits will be automatically reflected. So things would stay clean :)

pkg/validate/metadata_validate.go
```diff
@@ -0,0 +1,28 @@
+package validate
+
+import "fmt"
+
+func ValidateDefaultChannel(metabundle *MetaBundle) error {
```
sugarraysam, line 5: I would move all validators in a sub-package called `validators/` to split the main code, from the actual validators. 

Also I would have 1 validator per file. So suggesting following directory structure, where `ValidateDefaultChannel` exists in `default_channel.go`:

```bash
pkg
├── utils
│   ├── colors.go
│   └── error.go
└── validate
    ├── print.go
    ├── validate.go
    └── validators
        └── default_channel.go
```

pkg/validate/validate.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+}
+
+func (mb *MetaBundle) Validate(runMeta bool) []error {
+	errs := []error{}
+
+	if runMeta {
+		validators := GetAllMetaValidators()
+
+		printMetaHeading()
```
sugarraysam, line 34: [nit] Should we use logger instead of print to stdout? Thinking about when this code will run inside of a `ValidationWebhook`. Or we detect this somehow, and JSON log if running inside a `ValidationWebhook`, to better integrate /w fluentd/elasticsearch etc. This is a problem for future us probably.

pkg/validate/validate.go
```diff
@@ -0,0 +1,49 @@
...
+}
+
+// TODO: This will return a MetaBundle with corresponding bundle
+func NewMetaBundle(addonMeta *v1alpha1.AddonMetadataSpec) *MetaBundle {
+	return &MetaBundle{
+		AddonMeta: addonMeta,
+	}
+}
+
+func (mb *MetaBundle) Validate(runMeta bool) []error {
```
sugarraysam, line 28: I think we can drop the `runMeta` option here. We can run all validators globally, and if some of them want to ignore `Bundle` then they can just do so in the func implementation? 

WDYT?

pkg/validate/metadata_validate.go
```diff
@@ -0,0 +1,28 @@
...
+		}
+	}
+
+	if !valid {
+		return fmt.Errorf("could not find defaultChannel in channels")
+	}
+	return nil
+}
+
+func GetAllMetaValidators() []Validator {
```
sugarraysam, line 21: would move to `validate.go` and rename to `GetAllValidators()` as we have a single global validator concept.

pkg/validate/validate.go
yashvardhan-kukreja: This is a really good point. I agree with having a standardized log format to be propagated in this project.  We can make use of logrus for example, it provides JSON formatted logging out of the box. Or we can use any existing logging library (logger/logr iirc) which is already bootstrapped by kubebuilder but would have confirm once about the extent of their formatting capabilities.

pkg/validators/validator_addon_label.go
```diff
@@ -0,0 +1,21 @@
+package validators
+
+import (
+	"fmt"
+
+	"github.com/mt-sre/addon-metadata-operator/pkg/utils"
+)
+
+// TODO: preferable, have return type as a (bool, error) instead
```
sugarraysam, line 9: can remove this comment

Maybe add some doc for this validators?

```go
// ValidateAddonLabel - <validator description>
```

pkg/validators/validator_default_channel.go
```diff
@@ -0,0 +1,17 @@
+package validators
+
+import (
+	"fmt"
+
+	"github.com/mt-sre/addon-metadata-operator/pkg/utils"
+)
+
+func ValidateDefaultChannel(metabundle *utils.MetaBundle) (bool, error) {
```
sugarraysam, line 9: Lets add validator docs please

internal/cmd/validate.go
```diff
@@ -10,16 +10,28 @@ import (
 
 	addonsv1alpha1 "github.com/mt-sre/addon-metadata-operator/api/v1alpha1"
 	"github.com/mt-sre/addon-metadata-operator/pkg/utils"
+	"github.com/mt-sre/addon-metadata-operator/pkg/validate"
 	log "github.com/sirupsen/logrus"
 
 	"github.com/spf13/cobra"
 )
 
+type validateOptions struct {
```
sugarraysam, line 19: are we keeping these flags after all?

sugarraysam (commented):
Leaving few more comments. Good work everyone !

PR #7, opened 2021-10-28
Bundle validation boilerplate

ashishmax31:
Adds the required boilerplate to extract and parse the indexImage and perform validations on them.


yashvardhan-kukreja:
> Awesome work, great design very clean.
> 
> How do we want to do for testing? I would vote for integration tests, for each validator. And ideally we would cover both a valid case and a non-valid case.
> 
> The testing framework could be tackled in another PR as it will be quite involved. What I would do is:
> 
> 1. Create a plain `metaBundle` testing object, that is valid on all points (inside `testutil/`)
> 2. Create a `<validor_name>_test.go` file for each validator
> 3. Get the valid `metaBundle` make sure validator func passes
> 4. Get the valid `metaBundle` tweak few fields and make validator fail
> 
> Here created a new subtask: https://issues.redhat.com/browse/MTSRE-284

First prototype:
- https://github.com/mt-sre/addon-metadata-operator/pull/15

yashvardhan-kukreja (approved):
Great work :D

PR #15, opened 2021-11-01
MTSRE-284 | Prototype: validators' test suite

yashvardhan-kukreja:

To run the validators' tests:
```sh
make test-validators
```

## Background:

Every validator test is represented by a dedicated type (struct) which implements the `Validator` interface (`pkg/validate/validate_test.go`)

```go
type Validator interface {
	Name() string
	Run(utils.MetaBundle) (bool, error)
	SucceedingCandidates() []utils.MetaBundle
	FailingCandidates() []utils.MetaBundle
}
```

For implementing that interface, the Validator test's struct must have four methods:
- `Name() string` : Returns the name of the validator to which the test corresponds.
- `SucceedingCandidates() []utils.MetaBundle`: Returns a slice of `MetaBundle` where each element can successfully pass the validator.
- `FailingCandidates() []utils.MetaBundle`: Returns a slice of `MetaBundle` where each element fails the validator due to any reason.
- `Run(utils.MetaBundle) (bool, error)`:  Represents running the validator over a `MetaBundle` and returning a `bool` (success: true/false) and `error` (error while executing the validator) accordingly. Succeeding Candidate `MetaBundle` fed to this method will be expected to return `(true, nil)` and Failing Candidate `MetaBundle` fed to this method will be expected to return `(false, nil/non-nil)`. 
<details>
  <summary>For example (Click to expand): `pkg/validators/test_validator_default_channel.go`</summary>
  
  ```go
	package validators

	import (
		"github.com/mt-sre/addon-metadata-operator/api/v1alpha1"
		"github.com/mt-sre/addon-metadata-operator/pkg/utils"
	)

	type ValidatorDefaultChannelTestBundle struct{}

	func (val ValidatorDefaultChannelTestBundle) Name() string {
		return "Addon Default Channel Validator"
	}

	func (val ValidatorDefaultChannelTestBundle) Run(mb utils.MetaBundle) (bool, error) {
		return ValidateDefaultChannel(&mb)
	}

	func (val ValidatorDefaultChannelTestBundle) SucceedingCandidates() []utils.MetaBundle {
		return []utils.MetaBundle{
			{
				AddonMeta: &v1alpha1.AddonMetadataSpec{
					ID:             "random-operator",
					DefaultChannel: "alpha",
					Channels: []v1alpha1.Channel{
						{
							Name: "alpha",
						},
						{
							Name: "sigma",
						},
					},
				},
			},
			{
				AddonMeta: &v1alpha1.AddonMetadataSpec{
					ID:             "random-operator",
					DefaultChannel: "beta",
					Channels: []v1alpha1.Channel{
						{
							Name: "alpha",
						},
						{
							Name: "beta",
						},
					},
				},
			},
		}
	}

	func (val ValidatorDefaultChannelTestBundle) FailingCandidates() []utils.MetaBundle {
		return []utils.MetaBundle{
			{
				AddonMeta: &v1alpha1.AddonMetadataSpec{
					ID:             "random-operator",
					DefaultChannel: "alpha",
					Channels: []v1alpha1.Channel{
						{
							Name: "beta",
						},
						{
							Name: "sigma",
						},
					},
				},
			},
			{
				AddonMeta: &v1alpha1.AddonMetadataSpec{
					ID:             "random-operator",
					DefaultChannel: "beta",
					Channels: []v1alpha1.Channel{
						{
							Name: "alpha",
						},
					},
				},
			},
		}
	}
  ```
</details>

## How to add a new validator test

Say, you created a new validator called `validator_default_channel` under the file `pkg/validators/validator_default_channel.go` accordingly.

* Now, go ahead and create a new file `pkg/validators/test_validator_default_channel.go`.
* Inside this file, define a struct corresponding to this validator. It can be an empty struct or non-empty depending on the way you want to implement the testing of it. For example: `type ValidateDefaultChannelTestBundle struct {}`
* Now, make this struct implement the `Validator` interface defined under `pkg/validate/validate_test.go`. (Refer to the previous section for the explanation for these methods).
* After this, your tests are ready but you have to register them to the validators' test suite now.
* For that, proceed to `pkg/validate/validate_test.go` and look for `var validatorsToTest`.
* Under it, just append the struct corresponding to the validator-test you defined under `pkg/validators/test_validator_default_channel.go`. For example, in our case, it would look like this:
```go
...
// register the validators to test here by appending them to the following slice
var validatorsToTest []Validator = []Validator{
    ValidatorAddonLabelTestBundle{},  // ignore this, it was previously present
    ValidatorDefaultChannelTestBundle{}, // pay attention! we just added this
}
...
```
* That's it! Rest will be taken care of assuming that the methods you defined for the struct are aptly coded.
* Go ahead and run the tests by entering `make test-validators` and it will run:
```sh
╰─ make test-validators

[DEBUG] Go version used is go version go1.17.1 darwin/amd64
=== RUN   Test_AllValidators
=== RUN   Test_AllValidators/Addon_Label_Validator
=== PAUSE Test_AllValidators/Addon_Label_Validator
=== RUN   Test_AllValidators/Addon_Default_Channel_Validator
=== PAUSE Test_AllValidators/Addon_Default_Channel_Validator
=== CONT  Test_AllValidators/Addon_Label_Validator
=== CONT  Test_AllValidators/Addon_Default_Channel_Validator
--- PASS: Test_AllValidators (0.00s)
    --- PASS: Test_AllValidators/Addon_Label_Validator (0.00s)
    --- PASS: Test_AllValidators/Addon_Default_Channel_Validator (0.00s)
PASS
ok      github.com/mt-sre/addon-metadata-operator/internal/testutil     0.641s
```


internal/testutil/all_validators_test.go
```diff
@@ -0,0 +1,47 @@
...
+
+type Validator interface {
+	Name() string
+	Run(utils.MetaBundle) (bool, error)
+	SucceedingCandidates() []utils.MetaBundle
+	FailingCandidates() []utils.MetaBundle
+}
+
+// register the validators to test here by appending them to the following slice
+var (
```
sugarraysam, line 19: nit: do we need the extra `( )` ?

internal/testutil/all_validators_test.go
```diff
@@ -0,0 +1,47 @@
...
+
+import (
+	"testing"
+
+	"github.com/mt-sre/addon-metadata-operator/pkg/utils"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+type Validator interface {
```
sugarraysam, line 11: :heart: love interfaces

internal/testutil/all_validators_test.go
```diff
@@ -0,0 +1,47 @@
...
+	}
+)
+
+func Test_AllValidators(t *testing.T) {
+	for _, validator := range validatorsToTest {
+		validator := validator
+		t.Run(validator.Name(), func(t *testing.T) {
+			t.Parallel()
+			// testing the succeeding candidates
+			succeedingMetaBundles := validator.SucceedingCandidates()
```
sugarraysam, line 32: We could thread this even more aggressively. Since goroutines are pretty small constructs, it would probably provide a small performance increase (although we should measure it). This is prematured optimization but seems quite straightforward. Since we can use the testing framework to report errors it's also ultra simple and we dont need to use an error channel or anything like that.

```go
succeedingMetaBundles := validator.SucceedingCandidates()
failingMetaBundles := validator.FailingCandidates()

var wg sync.WaitGroup
wg.Add(len(succeedingMetaBundles) + len(failingMetaBundles))

for _, mb := range succeedingMetaBundles {
  go func(v Validator, mb utils.MetaBundle) {
    success, err := v.Run(mb)
    require.NoError(t, err)
    assert.True(t, success)
    wg.Done()
  }(validator, mb)
}

for _, mb := range failingMetaBundles {
  go func(v Validator, mb utils.MetaBundle) {
    success, err := v.Run(mb)
    require.NoError(t, err)
    assert.False(t, success)
    wg.Done()
  }(validator, mb)
}
wg.Wait()
```

sugarraysam (commented):
Good work I like this design. Few points we could improve but don't need to be added here but nice-to-have moving forward:
- `testutil/` helper methods to get a valid AddonMeta, with Bundles and Metadata set, from which we can only tweak a few fields, instead of recreating a new object.
- An example of how to include an actual bundle object in a `SucceedingCandidate` or `FailingCandidate`, probably with `testutil/` helper method as well

Would it make sense to make the following organization change ? Would be a little less cryptic IMO.
- move `internal/testutil/all_validators_test.go` to `pkg/validate/validate_test.go`
- move validator test files alongside their implementation:
  - `internal/testutil/validate_current_csv.go` -> `pkg/validators/validate_current_csv_test.go` 
  - `internal/testutil/validator_default_channel.go` -> `pkg/validators/validator_default_channel_test.go` 

Good PR description, this could even be added for developer instructions about the testing framework. Could be added inline in the code, or in `docs/testing.md` or `README.md`. Whatever you think is best!

internal/testutil/all_validators_test.go
sugarraysam: My only question would be, if one of the goroutine fails, are the other goroutines stopped, garbage collected? We would want an early exit.

yashvardhan-kukreja: Yep thought of this too but realised that the computation of these tests (at least until now) isn't I/O bound/blocking. They are simply CPU-bound non-blocking tasks, so I don't think that "cumulatively" the use of goroutines endup providing any performance benefits. 
Also, if in the future, our piece of code ends up generating, say, a 500 goroutines and if all of them are non-blocking only, then, they would actually end up imposing an increase in latency as well considering the context switches between 500 goroutines. Although the latency increase with N goroutines wouldn't be as substantial as when faced with N number of OS threads considering that the goroutines end-up multiplexing to one OS thread.

Also, the `t.Parallel()` is achieving a baseline parallelism/concurrency at a per-validator-test basis. The concurrency at a per-candidate basis seems a bit overkill to me and again, prone to spinning up a heck lot of goroutines and potential latency jumps.

That's what I think, would love to know your thoughts as well and would love to be corrected if I am wrong anywhere :)

internal/testutil/all_validators_test.go
yashvardhan-kukreja: ahh, old habits, die hard! Would clean this up :)

internal/testutil/all_validators_test.go
yashvardhan-kukreja: In case, we decide to proceed with implementing goroutines:

> My only question would be, if one of the goroutine fails, are the other goroutines stopped, garbage collected? We would want an early exit.

Regarding this, we can have an unbuffered channel called `yeet` (whatever). And the sender to that channel would be the piece of code executing the validator tests. Whenever a validator test fails, a value will be sent to this `yeet` channel.
The receiving end of this channel would be simply a separate goroutine which will be running an infinite loop and constantly checking if it received any value from the `yeet` channel. And whenever it will receive a value from the `yeet` channel, it will just `os.Exit(1)` our program.

So, in simple words, the flow will be:
- a common unbuffered channel of communication.
- Our validator tests run and the first time they encounter a failure, the send a value to that channel.
- In real-time, the receiver goroutine captures a correspondingly from that channel and takes it as an indication to run `os.Exit(1)`

PS: the only thing I want to confirm about is the garbage collection of other goroutines which will be under execution, when os.Exit(1) will happen

ashishmax31: >the only thing I want to confirm about is the garbage collection of other goroutines which will be under execution, when os.Exit(1) will happen
There is no garbage collection to be done after you call `os.Exit()` the process itself exits.

ashishmax31: @sugarraysam  I also agree with @yashvardhan-kukreja, I think scaling the number of go-routines linearly with the amount of validator tests to run wouldn't give us any extra performance benefits. I think the best approach here is to stick with `t.Parallel()`.

internal/testutil/validator_addon_label.go
```diff
@@ -0,0 +1,45 @@
...
+)
+
+type ValidatorAddonLabelTestBundle struct{}
+
+func (val ValidatorAddonLabelTestBundle) Name() string {
+	return "Addon Label Validator"
+}
+
+func (val ValidatorAddonLabelTestBundle) Run(mb utils.MetaBundle) (bool, error) {
+	return validators.ValidateAddonLabel(&mb)
```
ashishmax31, line 16: Unrelated to this PR, but I noticed that we're passing pointers to the validation fns. I think thats unnecessary and could cause problems with other validations, if someone accidentally modifies the metabundle object. Each validation fn should receive its own private copy of the metabundle object.

internal/testutil/all_validators_test.go
sugarraysam: Perfect. Thanks for your answer. I agree it is overkill for now. Let's revisit if there is ever a performance increase need.

internal/testutil/validator_addon_label.go
yashvardhan-kukreja: Bingo! Even I thought so while coding out this PR and have already a dedicated branch/commit ready locally shifting from pointer semantics to value semantics in the validator functions :)

PS: Nice catch though, even I prefer complying with value semantics instead of pointer semantics wherever possible in a consistent manner as pointer receivers (stored on heap) end up imposing more pressure on the GC as compared to value receivers (stored on stack) 🚀

yashvardhan-kukreja:
> Would it make sense to make the following organization change ? Would be a little less cryptic IMO.
> 
> move internal/testutil/all_validators_test.go to pkg/validate/validate_test.go
> move validator test files alongside their implementation:
> internal/testutil/validate_current_csv.go -> pkg/validators/validate_current_csv_test.go
> internal/testutil/validator_default_channel.go -> pkg/validators/validator_default_channel_test.go

@sugarraysam Nice suggestion, seems pretty conventional considering that usually, Go projects seems to place unit tests together with the business logic under same directories. And even would be slightly more convenient in the eyes of the new dev/engineer as they wouldn't have to `cd` into the `internal/testutil/` directory and deal with a different directory than `pkg/validators` to write tests 😛

yashvardhan-kukreja:
> Good PR description, this could even be added for developer instructions about the testing framework. Could be added inline in the code, or in docs/testing.md or README.md. Whatever you think is best!

@sugarraysam added the docs as well under `docs/VALIDATOR_TESTS.md` :)

internal/testutil/validator_addon_label.go
yashvardhan-kukreja: https://github.com/mt-sre/addon-metadata-operator/pull/17

Makefile
```diff
@@ -133,3 +133,6 @@ endef
 
 help: ## Display this help.
 	@awk 'BEGIN {FS = ":.*##"; printf "\nUsage:\n  make \033[36m<target>\033[0m\n"} /^[a-zA-Z_0-9-]+:.*?##/ { printf "  \033[36m%-15s\033[0m %s\n", $$1, $$2 } /^##@/ { printf "\n\033[1m%s\033[0m\n", substr($$0, 5) } ' $(MAKEFILE_LIST)
+
+test-validators:
```
sugarraysam, line 137: We can drop this target I think. Let's modify `make test` as needed, since this is what `pr_check.sh` will be running. Right now it is `$ go test -count=1 $(PKGS)`

docs/VALIDATOR_TESTS.md
```diff
@@ -0,0 +1,194 @@
...
+...
+// register the validators to test here by appending them to the following slice
+var validatorsToTest []Validator = []Validator{
+    ValidatorAddonLabelTestBundle{},  // ignore this, it was previously present
+    ValidatorDefaultChannelTestBundle{}, // pay attention! we just added this
+}
+...
+```
+* That's it! Rest will be taken care of assuming that the methods you defined for the struct are aptly coded.
+* Go ahead and run the tests by entering `make test-validators` and it will run:
```
sugarraysam, line 51: would change this line `make test-validators` as this target is not ran in `pr_check.sh`

docs/VALIDATOR_TESTS.md
```diff
@@ -0,0 +1,194 @@
...
+```sh
+╰─ make test-validators
```
sugarraysam, line 53: nit

docs/VALIDATOR_TESTS.md
```diff
@@ -0,0 +1,194 @@
...
+...
+// register the validators to test here by appending them to the following slice
+var validatorsToTest []Validator = []Validator{
+    ValidatorAddonLabelTestBundle{},  // ignore this, it was previously present
+    ValidatorDefaultChannelTestBundle{}, // pay attention! we just appended this
+}
+...
+```
+
+* That's it! Run `make test-validators` to see it in action :)
```
sugarraysam, line 176: nit

docs/VALIDATOR_TESTS.md
```diff
@@ -0,0 +1,194 @@
...
+
+```sh
+╰─ make test-validators
```
sugarraysam, line 179: nit

sugarraysam (approved):
Nice work. Let's merge this. Added few nits.

Let's keep those two points for a next PR, as nice to haves:
- `testutil/` helper method(s) to get a valid `MetaBundle`, with Bundles and Metadata set, from which we can only tweak a few fields, instead of recreating a new object.
- An example of how to include an actual bundle object in a `SucceedingCandidate` or `FailingCandidate`, probably with `testutil/` helper method(s) as well

Makefile
yashvardhan-kukreja: Ohh gotcha! Purging this, thanks :)

PR #18, opened 2021-11-12
Fixes: CGO_ENABLED=1, disable cross-compilation, fix cache path, os.Exit(1).

sugarraysam:
Couple of fixes from testing the binary.

## Fix cache path

`/tmp/mtcli` is used by sretoolbox to extract the Mtcli binary archive, adding a uuid to make sure we don't have collisions.

## os.Exit(1)

Exit code if validation fails was 0, want to detect in CI to do early exit.

## CGO_ENABLED=1

This is because one of the dependencies uses `go-sqlite3`:

```bash
$ rg CGO_ENABLED vendor
vendor/github.com/mattn/go-sqlite3/README.md
55:***Important: because this is a `CGO` enabled package you are required to set the environment variable `CGO_ENABLED=1` and have a `gcc` compile present within your path.***
161:This package requires `CGO_ENABLED=1` ennvironment variable if not set by default, and the presence of the `gcc` compiler.
182:    CGO_ENABLED=1 GOOS=linux GOARCH=arm GOARM=7 \
```

Need to modify `.goreleaser.yml` to also disable cross compilation (Windows + Darwin). We would need to run build on Github Darwin/Windows VM.

Tested new binary against reference addon:

```bash
$ ./bin/mtcli validate ../managed-tenants/addons/reference-addon/metadata/stage/addon.yaml

======RUNNING METADATA VALIDATORS======

Ensure defaultChannel is present in list of channels            Success
Ensure `label` to follow the format api.openshift.com/addon-<operator-id>               Success
Ensure current csv is present in the index image                Failed
```


sandhyadalavi (approved):
lgtm

PR #19, opened 2021-11-12
Run goreleaser inside a container for gcc bindings.

sugarraysam:


yashvardhan-kukreja:
reviewing this

build_tag.sh
```diff
@@ -2,27 +2,14 @@
...
-# uses config from .goreleaser.yml
-curl -sL https://git.io/goreleaser | bash
+# Run build_tag.sh in a sandbox to enforce golang & gcc bindings (CGO_ENABLED=1):
+# - sync secrets from app-interface:
+#   -> secrets: /resources/jenkins/global/secrets.yaml
+#   -> for job `gl-build-tag`: /resources/jenkins/global/templates.yaml
+IMAGE_TEST=addon-metadata-operator
+docker build -t ${IMAGE_TEST} -f Dockerfile.ci .
+docker run --rm \
+    -e "VERSION=" \
```
yashvardhan-kukreja, line 12: shouldn't we pin to a specific goreleaser version instead of relying on the latest version to be picked and used? (might coz issues in the future in case of latest version points to a major breaking release)

Dockerfile.build
```diff
@@ -2,13 +2,16 @@
 # --- stage 1 ---
 FROM registry.ci.openshift.org/openshift/release:golang-1.16 as builder
 
+ENV GO111MODULE=on
```
yashvardhan-kukreja, line 5:
```suggestion
```
can be removed as this is the default behaviour anyway with Go 1.16 :)

Dockerfile.build
```diff
@@ -2,13 +2,16 @@
...
+
 WORKDIR /build
 
 RUN echo "noroot:x:10001:10001:noroot:/:/sbin/nologin" | tee -a /etc/passwd
 
 # Cache optimization
-COPY go.mod go.sum ./
-RUN go mod download
+COPY go.mod go.sum Makefile ./
+COPY vendor ./vendor
```
yashvardhan-kukreja, line 13:
```suggestion
```
can be removed as the `make vendor` instruction in the next line brings back the vendor files anyway :)

Dockerfile.ci
```diff
@@ -0,0 +1,23 @@
+# vi: set ft=dockerfile:
+FROM registry.ci.openshift.org/openshift/release:golang-1.16
+
+ENV GO111MODULE=on
```
yashvardhan-kukreja, line 4:
```suggestion
```
duplicate: https://github.com/mt-sre/addon-metadata-operator/pull/19/files#r750124577 :)

Dockerfile.ci
```diff
@@ -0,0 +1,23 @@
...
+
+# Need latest version of git for goreleaser
+# https://packages.endpoint.com/
+RUN yum -y remove git \
+    && yum -y install https://packages.endpoint.com/rhel/7/os/x86_64/endpoint-repo-1.9-1.x86_64.rpm \
+    && yum -y install git
+
+# Cache optimization
+COPY go.mod go.sum Makefile ./
+COPY vendor ./vendor
```
yashvardhan-kukreja, line 18:
```suggestion
```
duplicate: https://github.com/mt-sre/addon-metadata-operator/pull/19/files#r750126268

Dockerfile.build
sugarraysam: done

Dockerfile.ci
sugarraysam: done

Dockerfile.build
sugarraysam: I thought it would speed up if we copy instead of redownloading?

As a side note, I saw they removed the `vendor/` dir in the `addon-operator` all together. Should we do the same? I feel since go modules this is not really required, and mostly annoying (especially in commit history). WDYT?

build_tag.sh
sugarraysam: Good point. Setting this.

Dockerfile.ci
sugarraysam: removed

Dockerfile.build
yashvardhan-kukreja: I am always up for getting rid of vendor/ , messes up a lot of PRs. Although I can think of only a very few (1-2) benefits of vendoring, it seems like an overkill to me. So yeah! let's purge it \o/

Regarding addon-operator, some context: we added some submodules underneath it and for making that compliant and not end up vendoring our own repo, we purged vendoring.

yashvardhan-kukreja (approved):

PR #21, opened 2021-11-18
list-bundles subcommand

ashishmax31:
This PR adds a new subcommand `list-bundles` to list the operator bundles present in an index image

### DIFF
--- a/internal/cmd/list_bundles.go
+++ b/internal/cmd/list_bundles.go
@@ -22,7 +22,7 @@ var (
 	listBundlesCmd = &cobra.Command{
 		Use:     "list-bundles",
 		Short:   "List all the bundles present in an index image.",
-		Example: strings.Join(validateExamples, "\n"),
+		Example: strings.Join(listBundlesExamples, "\n"),
 		Args:    cobra.ExactArgs(1),
 		Run:     listBundlesMain,
 	}
--- a/internal/cmd/list_bundles.go
+++ b/internal/cmd/list_bundles.go
@@ -30,16 +30,18 @@ var (
 
 func listBundlesMain(cmd *cobra.Command, args []string) {
 	indexImageUrl := args[0]
-	allBundles, err := utils.ExtractAndParseAllAddons(indexImageUrl)
+	allBundles, err := utils.ExtractAndParseAddons(indexImageUrl, utils.AllAddonsIdentifier)
 	if err != nil {
 		log.Fatalf("Failed to extract and parse bundles from the given index image: Error: %s \n", err.Error())
 	}
-
+	var operatorVersionedNames []string
 	for _, bundle := range allBundles {
 		version, err := bundle.Version()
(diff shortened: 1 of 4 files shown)


mayankshah1607:
Will this command be called within the `managed-tenants-bunldes` CI to validate against the bundles directory?

ashishmax31:
@mayankshah1607 Yes

internal/cmd/list_bundles.go
```diff
@@ -0,0 +1,45 @@
...
+)
+
+func init() {
+	mtcli.AddCommand(listBundlesCmd)
+}
+
+var (
+	listBundlesExamples = []string{
+		"  #List all the bundles present in an index image.",
+		"  mtcli list-bundles <index_image>",
```
sugarraysam, line 20: It seems if we provide the `--addon-name` flag, it would simplify the code so much. Also, since this is the cache key, it should help with performance.

If I understand correctly, by providing `--addon-name` we could remove the new function altogether `ExtractAndParseAllAddons` and load manifests based on the `addonName` instead of iterating over all operators?

I think it's a reasonable ask, tenants will for sure know the addon-name when using `$ mtcli list-bundles` and for our use case, which is to validate index images before pushing, we will also know the addon-name.

WDYT?

Otherwise it seems I don't see the difference between `ExtractAndParseAddon` and the new `ExtractAndParseAllAddons`?

ashishmax31: @sugarraysam The issue with that approach is that indexImages can come with other dependent operators, like the ocs-addon. So it'll be very awkward to type in the names of the addon + all the dependent addons.

ashishmax31: Yep, makes sense.

ashishmax31:
@sugarraysam  Have a look now :)

pkg/utils/bundle_utils.go
```diff
@@ -2,38 +2,87 @@ package utils
...
+		cacheDir:     defaultCacheDir,
+		indexImage:   indexImage,
+	}
+
+	key := indexImageExtractor.CacheKey(indexImage, addonIdentifier)
+	if err := extractAddons(indexImageExtractor, key, indexImage); err != nil {
+		return []registry.Bundle{}, err
+	}
+	if addonIdentifier == AllAddonsIdentifier {
+		// Parse all adddons in the extraction path
```
sugarraysam, line 37: Extra comment here help alot :+1:

sugarraysam (approved):
Thank you for this work. I found this new version much easier to understand, and the two added extra comments quite handy.

Good job !

ashishmax31:
@sugarraysam I'll add the integration tests as part of a separate PR.

ashishmax31:
@sugarraysam Added int tests in this PR https://github.com/mt-sre/addon-metadata-operator/pull/22

PR #22, opened 2021-11-22
Add integration tests for mtcli list-bundles cmd

ashishmax31:


Makefile
```diff
@@ -23,6 +24,11 @@ all: build
...
 	@go test -count=1 $(PKGS)
 
+
+test-e2e: ## Run e2e integration tests
+	@CGO_ENABLED=1 go build -mod=vendor -a -o $(E2E_MTCLI_PATH) cmd/mtcli/main.go
+	@E2E_MTCLI_PATH=$(E2E_MTCLI_PATH) go test $(INTEGRATION_TESTS)
+
 check: golangci-lint goimports ## Runs all checks.
 	@go fmt $(PKGS)
 	@go vet $(PKGS)
```
yashvardhan-kukreja, line 34:
```suggestion
check: golangci-lint goimports ## Runs all checks.
	@go fmt $(PKGS) $(INTEGRATION_TESTS)
	@go vet $(PKGS) $(INTEGRATION_TESTS)
```
Minor nit :)

integration/mtcli/list_bundles_test.go
```diff
@@ -0,0 +1,106 @@
...
+	s.Require().NoError(err, "Cant find mtcli binary at E2E_MTCLI_PATH")
+}
+
+func (s *e2eTestSuite) TestMtcliListBundlesE2E() {
+	type testCaseArg struct {
+		indexImageUrl  string
+		expectedOutput []string
+	}
+
+	testCases := []testCaseArg{
```
yashvardhan-kukreja, line 45:
```suggestion
	testCases := []struct{
	        indexImageUrl  string
		expectedOutput []string
	}{
```
Seems like a cleaner place to use anonymous structs from my eyes. WDYT :)

yashvardhan-kukreja:
Good job @ashishmax31 !
Just some tiny aforementioned suggestions, rest all LGTM :)

yashvardhan-kukreja (approved):

Makefile
```diff
@@ -23,9 +24,14 @@ all: build
 test: ## Run tests.
 	@go test -count=1 $(PKGS)
 
+
+test-e2e: ## Run e2e integration tests
```
sugarraysam, line 28: I am a huge fan of testing and integration test, and think it's great work, but I am wondering if this is worth the extra effort of building and running the binary, when you could test the `listBundlesMain(...)` function without any extra toil/preparation? I don't see what we are gaining from running integration tests here vs. regular unit tests of the `listBundlesMain(...)` function

integration/mtcli/list_bundles_test.go
```diff
@@ -0,0 +1,104 @@
...
+				"ocs-osd-deployer.v1.0.0",
+				"ocs-osd-deployer.v1.0.1",
+				"ocs-osd-deployer.v1.0.2",
+				"ocs-osd-deployer.v1.1.0",
+				"ocs-osd-deployer.v1.1.1",
+				"ose-prometheus-operator.v4.8.0",
+			},
+		},
+	}
+	for _, testCase := range testCases {
```
sugarraysam, line 67: Can use `t.Run(...) + t.Parallel(...)` to speedup tests.

Makefile
ashishmax31: I think we should have separate unit tests along with integration tests. The idea of this integration test(and future ones) would be to test things from the outside(like how our users would use it) and make sure that the interface we give out to our clients is proper.(Stuff like proper formatting of output, how we take in input through STDIN etc). While I agree that this feature right now is too trivial, the idea still stands.

Makefile
```diff
@@ -23,9 +24,14 @@ all: build
...
+	@CGO_ENABLED=1 go build -mod=vendor -a -o $(E2E_MTCLI_PATH) cmd/mtcli/main.go
+	@E2E_MTCLI_PATH=$(E2E_MTCLI_PATH) go test $(INTEGRATION_TESTS)
+
 check: golangci-lint goimports ## Runs all checks.
-	@go fmt $(PKGS)
-	@go vet $(PKGS)
+	@go fmt $(PKGS) $(INTEGRATION_TESTS)
```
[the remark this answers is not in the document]
ashishmax31: Nope, i separated them out here: https://github.com/mt-sre/addon-metadata-operator/blob/9864ad0bdc74f997de00f3d66775af66aaa8c7f6/Makefile#L12

Makefile
```diff
@@ -23,9 +24,14 @@ all: build
...
+	@go vet $(PKGS) $(INTEGRATION_TESTS)
```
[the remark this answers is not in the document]
ashishmax31: Nope, i separated them out here:
https://github.com/mt-sre/addon-metadata-operator/blob/9864ad0bdc74f997de00f3d66775af66aaa8c7f6/Makefile#L12

integration/mtcli/list_bundles_test.go
ashishmax31: I don't think i can run it in parallel now, the stuff around caching is not thread safe at the moment (file might get overwritten and cause the assertions to fail)

ashishmax31: testify/suite doesn't support parallel test execution - https://github.com/stretchr/testify/issues/187

Makefile
sugarraysam: I am sold :) , thank you

Makefile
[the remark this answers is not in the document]
sugarraysam: my mistake !

integration/mtcli/list_bundles_test.go
sugarraysam: interesting, I did not know that

PR #23, opened 2021-11-26
change how operator name is constructed from a parsed bundle

ashishmax31:
This PR changes how operator names are constructed from a parsed bundle. Earlier, we were constructing the name from the package name + version. Turns out this is not ideal as some operators can have different package name compared to the name in the CSV.
Signed-off-by: ashish <asnaraya@redhat.com>


sugarraysam (approved):
lgtm

PR #26, opened 2021-12-01
Add a --disable/--enabled flag to validate cmd.

sugarraysam:
This is necessary and provides more flexibility when integrating with the managed-tenants pipeline.

We are currently facing an issue where validators are not passing checks, and this is a blocker. As we work on fixing those validators, we can temporarily disable them and keep working on new validators, as well as fixing older ones.

Will also be useful if we ever introduce a bug and somehow need to quickly pass the pipeline.

### DIFF
--- a/pkg/validate/validate.go
+++ b/pkg/validate/validate.go
@@ -1,21 +1,22 @@
 package validate
 
 import (
+	"errors"
 	"fmt"
+	"strings"
 
 	"github.com/mt-sre/addon-metadata-operator/pkg/validators"
 
 	"github.com/mt-sre/addon-metadata-operator/pkg/utils"
 )
 
-func Validate(mb utils.MetaBundle) (bool, []error) {
+func Validate(mb utils.MetaBundle, filter *validatorsFilter) (bool, []error) {
 	errs := []error{}
 	allSuccess := true
-	validators := GetAllValidators()
 
 	printMetaHeading()
 
-	for _, validator := range validators {
+	for _, validator := range filter.GetValidators() {
 		fmt.Printf("\r%s\t\t", validator.Description)
 		success, err := validator.Runner(mb)
 		if err != nil {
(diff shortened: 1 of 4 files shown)


sugarraysam:
Let's not merge right now, will also add an `--enabled` flag to facilitate testing a single validator against addons living in the managed-tenants repository.

pkg/validate/validate.go
```diff
@@ -32,19 +33,108 @@ func Validate(mb utils.MetaBundle) (bool, []error) {
...
+	var res []utils.Validator
+	for _, name := range f.ValidatorNames {
+		// can ignore _ok_ was already checked
+		validator, _ := AllValidators[name]
+		res = append(res, validator)
+	}
+	return res
+}
+
+func (f *validatorsFilter) getDisabledValidators() []utils.Validator {
```
yashvardhan-kukreja, line 125:
```suggestion
func (f *validatorsFilter) getNotDisabledValidators() []utils.Validator {
```
Or maybe we can change it some other more suitable name? `getDisabledValidators` sounds like something which would return validators which are disabled but in reality, the method is doing the opposite.
wdyt :)

pkg/validate/validate.go
```diff
@@ -32,19 +33,108 @@ func Validate(mb utils.MetaBundle) (bool, []error) {
...
+	var res []string
+	for name := range AllValidators {
+		res = append(res, name)
+	}
+	return strings.Join(res, ",")
+}
+
+type validatorsFilter struct {
+	ReturnAll      bool
+	Enabled        bool
```
yashvardhan-kukreja, line 62: Do we really need to track `Enabled`? :/
can't it be just 
```go
type validatorsFilter struct {
    ReturnAll bool // same meaning as before
    validatorNames []string // effective names of validators to finally consider (enabled validators / all minus disabled validators)
}
```

And the `NewFilter` function/constructor can have the following pseudocode
```
if disabled == "" and enabled == "" 
    return ValidatorsFilter{ReturnAll: true}

if enabled != "" 
    validatorNames := strings.Split(enabled, ",")
if disabled != ""
    validatorNames := strings.Split(allValidators minus disabled, ",")

return validatorsFilter {ReturnAll: false, validatorsNames: validatorNames}
```

With this we can easily modify the `GetValidators` function to this
```
output := []utils.Validator
if f.returnAll {
    for _, validator := range AllValidators {
        output = append(output, validator)    ​
    }
} else {
    for _, name := f.validatorNames {
        validator := AllValidators[name] // no need to track 'ok' as it was already validated in the NewFilter function
        output = append(output, validator)
    }
}

return output
```

With this, codebase would become much cleaner as `GetAllValidators()`, `GetEnabledValidators`, `GetDisabledValidators` would have gotten rid of.

Looks pretty simple and cleaner to me, wdyt :)

sugarraysam: I like this, let me refactor my code.

pkg/validate/validate.go
sugarraysam: refactored, this function is gone

yashvardhan-kukreja:
lgtm! great work, thanks :)

yashvardhan-kukreja (approved):
lgtm! great work, thanks :)

PR #27, opened 2021-12-02
MTSRE-330 | Get rid of vendoring

yashvardhan-kukreja:


yashvardhan-kukreja:
image

.pre-commit-config.yaml
```diff
@@ -5,21 +5,8 @@
 # Update hooks:
 #   $ pre-commit autoupdate
 ---
-exclude: |
-  (?x)(
-    (^vendor/)|
-    (.deepcopy.go$)
-  )
+exclude: ".*.deepcopy.go"
```
sugarraysam, line 8: Think we need to escape? But since most of our validators use `pass_filenames: false` this is ignored.

`.*\.deepcopy\.go$`

sugarraysam (approved):
approving, thanks for this. Missing few lines in makefile:
```makefile
PKGS := $(shell go list ./... | grep -v -E '/vendor|/integration')

<...>
@$(GOIMPORTS) -w -l $(shell find . -type f -name "*.go" -not -path "./vendor/*")

```

Dockerfile.ci
```diff
@@ -11,7 +11,7 @@ RUN curl -sL https://git.io/goreleaser -o /tmp/goreleaser.sh \
 
 # Cache optimization
 COPY go.mod go.sum Makefile ./
-RUN go mod download && make vendor
+RUN go mod download
```
sugarraysam, line 14: should we replace /w `make tidy`?

Dockerfile.build
```diff
@@ -10,7 +10,7 @@ RUN echo "noroot:x:10001:10001:noroot:/:/sbin/nologin" | tee -a /etc/passwd
 
 # Cache optimization
 COPY go.mod go.sum Makefile ./
-RUN go mod download && make vendor
+RUN go mod download
```
sugarraysam, line 13: `&& make tidy` ?

Dockerfile.ci
yashvardhan-kukreja: make tidy ends up running `go mod tidy` and the thing with go mod tidy is that it updates and then, downloads the modules listed in go.mod. The risk with that is that say, in runtime, `make tidy`/`go mod tidy` ends up upgrading one of our modules to a newer major breaking version, that would break our code and it would be hard to track it coz in the repository everything would seem to be working fine.

For reference, this is a similar concern with addon-operator. If you run `go mod tidy` over it, it will upgrade logr/go-logger from 0.4.0 to 1.2.0 which breaks the entire addon-operator.

although as of now, we don't have any module in go.mod, which even upgraded, would cause any breakages in runtime. So, we can afford to proceed with `make tidy` as of now but god knows what will happen in the future 🤞🏼 

wdyt :)

.pre-commit-config.yaml
yashvardhan-kukreja: ".*.deepcopy.go" works :) I tested it out manually locally by editing some deepcopy.go in dirty manner and committing them and pre-commit was ignoring them.

Same thing I recently confirmed and committed here as well - https://github.com/openshift/reference-addon/blob/main/.pre-commit-config.yaml#L1

Though, if you noticed some edge-case/issue with the above regex do let me know. I won't be surprised if I missed one as regexes are edgy too deal with 😂

sugarraysam: perfect.

Dockerfile.ci
sugarraysam: awesome let's keep it out.

.pre-commit-config.yaml
```diff
@@ -5,21 +5,8 @@
...
 repos:
-  # vendor
-  - repo: local
-    hooks:
-      - id: govendor
-        name: govendor
-        language: system
-        entry: make vendor
```
sugarraysam, line 20: Just noticed, can we replace this with `make tidy`?

sugarraysam (approved):

PR #31, opened 2021-12-09
Support dynamic typing in addOnRequirementData, fix suboperators snake case.

sugarraysam:
## Motivation

~~There is a discrepancy between OCM /versions endpoint (imageset) and the metadata endpoint where the `addOnParameters` can either have an `items` object or not. This PR modifies and fixes this issue so we can easily work with both scenarios.~~

## Update

Since the schema between metadata and imageset is now uniform. This PR makes sure that we can do:
- Dynamic typing in addOnRequirementData
- Fixes snake case in subOperators spec

## Testing

Compiled the binary locally, and tested against our production addons, as per this SOP: https://gitlab.cee.redhat.com/service/managed-tenants-sops/-/blob/main/MT-SRE/sops/AMO-mtcli.md#run-mtcli-validate-against-all-addons.


mayankshah1607:
Would it be easier if the ImageSet matched with the metadata schema? I.e: no .items field?
We'd have to modify the payload to OCM on the fly, something like this: https://github.com/mt-sre/managed-tenants-cli/blob/main/managedtenants/utils/ocm.py#L200

sugarraysam:
I think it would make more sense, to provide a consistent interface to our tenants. But either way works.

mayankshah1607:
> I think it would make more sense, to provide a consistent interface to our tenants. But either way works.

FWIW, we'd be discontinuing / deprecating the parameters, requirements, etc fields from metadata anyway once we have full migration to imagesets.

Let's hold make the imageset schema consistent with metadata. Meanwhile, let's put this one on hold.

mayankshah1607:
https://github.com/mt-sre/managed-tenants-cli/pull/66

sandhyadalavi (approved):
lgtm

PR #32, opened 2021-12-13
MTSRE-352 | synchronize addonmetadata api

yashvardhan-kukreja:
Signed-off-by: Yashvardhan Kukreja <ykukreja@redhat.com>

Synchronize `AddonMetadataSpec` and `ImageSetSpec` with [addon metadata schema](https://github.com/mt-sre/managed-tenants-cli/blob/main/managedtenants/data/metadata.schema.yaml) and [imageset schema](https://github.com/mt-sre/managed-tenants-cli/blob/main/managedtenants/data/imageset.schema.yaml)

PR #34, opened 2021-12-13
Fix validators interface. Add refAddonStage testutils. Fix 001_default_channel.

sugarraysam:
This PR adds a few changes and can be a bit confusing because many files are touched, but they all work together.

## Changes
- Add `testutils.GetReferenceAddonStage` to help writing validators, especially getting a quick and easy `SucceedingCandidate`, that includes a full `MetaBundle` object
- Fix the `001_default_channel` validator
- Modifies the validator interface so validators can return a `failureMsg` that will give users instructions on how to fix the failure
- Refactored the `meta_loader_test.go` to use the `testutils.GetReferenceAddonStage` helper

## Issue found with `&registry.Annotations{}`

It seems our extraction code extracts bundles using the deprecated `packageManifest` format:

```bash
# Inspect the local cache
$ tree /tmp/mtcli-07b10894-0673-4d95-b6ef-0cbd9701c9c3
/tmp/mtcli-07b10894-0673-4d95-b6ef-0cbd9701c9c3
└── quay.io
    └── osd-addons
        └── reference-addon-index@sha256:0c8b02008f2c2faeb681ae8cd454821266a794435aea4b3f7ae28c74bc2e280d
            └── reference-addon
                ├── 0.1.0
                │   └── reference-addon.csv.yaml
                ├── 0.1.1
                │   └── reference-addon.csv.yaml
                ├── 0.1.2
                │   └── reference-addon.csv.yaml
                ├── 0.1.3
                │   └── reference-addon.csv.yaml
                ├── 0.1.4
                │   └── reference-addon.csv.yaml
                ├── 0.1.5
                │   └── reference-addon.csv.yaml
                └── package.yaml
```
The issue with this format is that we do not get access to the `annotations.yaml` file. Now I might be wrong, but I believe we need to validate some of these annotations as OLM depends on them to run the workloads properly. For example, look at the commented out `pkg/validators/validator_001_default_channel.go::matchesDefaultChannelAnnotations::#line 61` which validates default channel annotations.

cc @yashvardhan-kukreja @ashishmax31

## Bundles format expected output (from `mtbundles`)

```bash
├── 0.1.0
│   ├── manifests
│   │   └── reference-addon.csv.yaml
│   └── metadata
│       └── annotations.yaml
├── 0.1.1
│   ├── manifests
│   │   └── reference-addon.csv.yaml
│   └── metadata
│       └── annotations.yaml
├── 0.1.2
│   ├── manifests
│   │   └── reference-addon.csv.yaml
│   └── metadata
│       └── annotations.yaml
├── 0.1.3
│   ├── manifests
│   │   └── reference-addon.csv.yaml
│   └── metadata
│       └── annotations.yaml
├── 0.1.4
│   ├── manifests
│   │   └── reference-addon.csv.yaml
│   └── metadata
│       └── annotations.yaml
├── 0.1.5
│   ├── manifests
│   │   └── reference-addon.csv.yaml
│   └── metadata
│       └── annotations.yaml
```


api/v1alpha1/helpers.go
```diff
@@ -11,3 +16,24 @@ func (a *AddonMetadataSpec) FromYAML(data []byte) error {
 func (a *AddonImageSetSpec) FromYAML(data []byte) error {
 	return yaml.Unmarshal(data, a)
 }
+
+// GetSemver - Returns the semver version matching "MAJOR.MINOR.PATCH".
+func (a *AddonImageSetSpec) GetSemver() (string, error) {
+	if parts := strings.SplitN(a.Name, ".", 2); len(parts) == 2 {
```
yashvardhan-kukreja, line 22: Isn't this function too permissive?
An input like "x.a.v.c.x.x." would pass too where instead it should be raising error right?

wdyt of this suggestion (validates the to-be-returned semver string as well)
```
import "golang.org/x/mod/semver"
```
```suggestion
	if parts := strings.SplitN(a.Name, ".", 2); len(parts) == 2 && semver.IsValid(parts[1]) {
```

api/v1alpha1/helpers_test.go
```diff
@@ -0,0 +1,33 @@
+package v1alpha1
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestGetSemver(t *testing.T) {
+	cases := []struct {
```
yashvardhan-kukreja, line 10: Let's add some test cases which end up returned expected errors when provided with improper input/semver strings?

pkg/validators/validate_current_csv.go
```diff
@@ -7,24 +7,25 @@ import (
 	"github.com/operator-framework/operator-registry/pkg/registry"
 )
 
-func ValidateCSVPresent(metabundle utils.MetaBundle) (bool, error) {
+// TODO - review failureMsg ++ missing test_validate_csv testing bundle
+func ValidateCSVPresent(metabundle utils.MetaBundle) (bool, string, error) {
 	if len(metabundle.Bundles) == 0 {
-		return false, fmt.Errorf("No bundles present!")
+		return false, "", fmt.Errorf("No bundles present!")
```
yashvardhan-kukreja, line 13:
```suggestion
		return false, "No bundles present", nil
```
Shouldn't it be this, because errors were meant to be indicative of unexpected bugs/errs ?

pkg/validators/validator_addon_label.go
```diff
@@ -5,11 +5,12 @@ import (
 )
 
 // ValidateAddonLabel validates whether the 'label' field under an addon.yaml follows the format 'api.openshift.com/addon-<id>'
-func ValidateAddonLabel(metabundle utils.MetaBundle) (bool, error) {
+// TODO - remove this validator once we do field level validation
+func ValidateAddonLabel(metabundle utils.MetaBundle) (bool, string, error) {
 	operatorId, label := metabundle.AddonMeta.ID, metabundle.AddonMeta.Label
 	if label != "api.openshift.com/addon-"+operatorId {
-		return false, nil
+		return false, "", nil
```
yashvardhan-kukreja, line 12:
```suggestion
		return false, "fmt.Sprintf("addon label '%s' wasn't recognised to follow the 'api.openshift.com/addon-<id>' format", label)", nil
```
how does this sound? :)

internal/testutils/reference_addon.go
```diff
@@ -0,0 +1,121 @@
...
+	Env            string
+}
+
+var (
+	ReferenceAddonImageSetDir   = path.Join(AddonsImagesetDir, "reference-addon")
+	ReferenceAddonIndexImageDir = path.Join(AddonsIndexImageDir, "reference-addon")
+	instance                    *ReferenceAddonStage
+)
+
+// GetReferenceAddonStage - uses singleton pattern to avoid loading yaml manifests over and over
```
yashvardhan-kukreja, line 37: Sweet!

pkg/validators/validator_001_default_channel.go
```diff
@@ -0,0 +1,91 @@
...
+		channelNames = append(channelNames, channel.Name)
+	}
+	return false, fmt.Sprintf("The defaultChannel '%v' is not part of the listed channelNames: %v.", defaultChannel, channelNames)
+}
+
+// TODO - (sblaisdo) enable when extraction format is bundles instead of packageManifest
+// if the annotation is not present, make sure the defaultChannel is alpha
+// func matchesDefaultChannelAnnotations(bundles []registry.Bundle, defaultChannel string) (bool, string, error) {
+// 	for _, bundle := range bundles {
+// 		allChannels := []string{bundle.Annotations.Channels, "dontexist"}
```
yashvardhan-kukreja, line 63:
```suggestion
//               if bundle.Annotations != nil {
//                   allChannels := append(bundle.Annotations.Channels.Split(","), "dontexist"}
//               }
```

- The `if Bundle.Annotations` because without a nil-check this might trigger a nil-dereference panic if `bundle.Annotations` doesn't exist. As `Annotations` is a ptr and behind the scenes, `bundle.Annotations.Channels` would be auto-dereferenced to access `Annotations.Channels`


- `allChannels := []string{bundle.Annotations.Channels.Split(",")..., "dontexist"}` because bundle.Annotations.Channels is a comma-separated string of all the channels. So, before including it in the `allChannels` slice, it should be split by comma and the resulting slice + "dontexist" should equate to `allChannels`

pkg/validators/validator_001_default_channel.go
```diff
@@ -0,0 +1,91 @@
...
+// 		var defaultBundleChannel string
+
+// 		if len(allChannels) == 1 {
+// 			// if a single channel is listed, should match defaultChannel
+// 			defaultBundleChannel = allChannels[0]
```
yashvardhan-kukreja, line 68: Shouldn't here be a sanity check to confirm that `allChannels[0]` / `defaultBundleChannel` matches `bundle.Annotations.DefaultChannelName` ? 

Or should there be a dedicated check for that?

pkg/validators/validator_001_default_channel.go
```diff
@@ -0,0 +1,91 @@
...
+// 				return false, "", fmt.Errorf("Could not read bundle version, got %v.", err)
+// 			}
+// 			msgs := []string{fmt.Sprintf("Missing or invalid channel annotation for bundle '%v:%v'", bundle.Name, version)}
+// 			msgs = append(msgs, fmt.Sprintf("DefaultChannel '%v' does not match channel from bundle annotations '%v'.", defaultChannel, defaultBundleChannel))
+// 			if len(allChannels) == 1 {
+// 				msgs = append(msgs, fmt.Sprintf("Please set the 'operators.operatorframework.io.bundle.channels.v1' annotation to '%v'.", defaultChannel))
+// 			} else {
+// 				msgs = append(msgs, fmt.Sprintf("Please set the operators.operatorframework.io.bundle.channel.default.v1 annotation to '%v'.", defaultChannel))
+// 			}
+// 			return false, strings.Join(msgs, "\n"), nil
```
yashvardhan-kukreja, line 87:
```suggestion
// 			return false, strings.Join(msgs, ": "), nil
```
Minor nit :)
(coz we're (and usually, as far as I have noticed) delimiting error/failure strings with ": ")

yashvardhan-kukreja:
Fab work @sugarraysam ! 
Just some minor suggestions and doubts. Rest we're good to go :)

ashishmax31:
@sugarraysam Interesting, let me go through the code today and find out why its exporting bundles in the package manifests format.

api/v1alpha1/helpers.go
sugarraysam: done

api/v1alpha1/helpers_test.go
sugarraysam: done

pkg/validators/validate_current_csv.go
sugarraysam: The way I understand our interface is this:
> // ValidateFunc - returns a triple consisting of:
> // 1. bool       - true if MetaBundle validation was successful
> // 2. failureMsg - "" if the result was true, else information about why the validation failed
> // 3. error      - report any error that happened in the validation code

sugarraysam: Basically, we want to distinguish code error from validation error.

pkg/validators/validator_addon_label.go
sugarraysam: fixed

pkg/validators/validator_001_default_channel.go
sugarraysam: oh I forgot to fix this, was just debugging stuff locally. Fixed the commented out code should be:

```go
allChannels := strings.Split(bundle.Annotations.Channels, ",")
```
```

File: pkg/validators/validator_001_default_channel.go (hunk 7 above)
Comment (sugarraysam) [reply pkg/validators/validator_001_default_channel.go:68]: So it's two cases:
1. if you have a single channel listed in `operators.operatorframework.io.bundle.channels.v1`, then you want the `defaultChannel` to match this
2. If you have multiple channels listed in `operators.operatorframework.io.bundle.channels.v1`, then you need the extra annotation `operators.operatorframework.io.bundle.channel.default.v1` matching the `defaultChannel`

File: pkg/validators/validator_001_default_channel.go (hunk 8 above)
Comment (sugarraysam) [reply pkg/validators/validator_001_default_channel.go:87]: I don't have a strong opinion on this. We just have to keep in mind those error msgs are printed both on stdout (CLI validation) and eventually in the validation-webhook-server response payload. Fixed!

File: pkg/validators/validate_current_csv.go (hunk 3 above)
Comment (sugarraysam) [reply pkg/validators/validate_current_csv.go:13]: EDIT:

you are right , fixed it ! Sorry brain fart

File: pkg/utils/bundle_utils_test.go [hunk 9]
```diff
@@ -1,15 +1,16 @@
-package utils
+package utils_test
```
sugarraysam, line 1: FYI: added this to avoid import cycle with utils -> <- testutils

api/v1alpha1/helpers.go
```diff
@@ -11,3 +18,45 @@ func (a *AddonMetadataSpec) FromYAML(data []byte) error {
...
+	if len(parts) == 2 {
+		version := parts[1]
+		if semver.IsValid(version) {
+			return strings.TrimPrefix(version, "v"), nil
+		}
+	}
+	return "", fmt.Errorf("Could not parse the imageSet name as a valid semver, %v.", a.Name)
+}
+
+func (a *AddonMetadataSpec) PatchWithImageSet(imageSet *AddonImageSetSpec) error {
```
sugarraysam, line 34: cc @yashvardhan-kukreja updated after rebasing #32

pkg/validators/validator_001_default_channel.go
```diff
@@ -0,0 +1,91 @@
...
+		channelNames = append(channelNames, channel.Name)
+	}
+	return false, fmt.Sprintf("The defaultChannel '%v' is not part of the listed channelNames: %v.", defaultChannel, channelNames)
+}
+
+// TODO - (sblaisdo) enable when extraction format is bundles instead of packageManifest
+// if the annotation is not present, make sure the defaultChannel is alpha
+// func matchesDefaultChannelAnnotations(bundles []registry.Bundle, defaultChannel string) (bool, string, error) {
+// 	for _, bundle := range bundles {
+// 		allChannels := strings.Split(bundle.Annotations.Channels, ",")
```
yashvardhan-kukreja, line 63:
```suggestion
//               if bundle.Annotations == nil {
//                   continue
//               }
// 		allChannels := strings.Split(bundle.Annotations.Channels, ",")
```

https://github.com/mt-sre/addon-metadata-operator/pull/34#discussion_r768602631

> The if Bundle.Annotations because without a nil-check this might trigger a nil-dereference panic if bundle.Annotations doesn't exist. As Annotations is a ptr and behind the scenes, bundle.Annotations.Channels would be auto-dereferenced to access Annotations.Channels
 
:)

pkg/validators/validator_001_default_channel.go
yashvardhan-kukreja: No I meant that say, there's only single channel under `allChannels`, say, it is "alpha". In that case, say, even the `bundle.Annotations.defaultChannel` is specified (although it's optional in this case), and say, it is "beta".

Then that's an incoherence right? And a validation error should be reported to either suggest to remove the `bundle.Annotations.defaultChannel` field because it is not needed as there's anyway only one channel under `bundle.Annotations.Channels` or to ensure that the `bundle.Annotations.defaultChannel` matches the single only channel under `bundle.Annotations.Channels`.

But I believe a separate validator for this check would be more suitable. Let me create an item for that in google sheet, create a JIRA ticket and wrap it up :)

api/v1alpha1/helpers.go
yashvardhan-kukreja: Great stuff!

pkg/validators/validator_001_default_channel.go
sugarraysam: good point. I fixed it, but we expect the extractor to fail if there are no annotations.yaml files, or if there is any problem reading it (not the case right now). I reflected that in the code.

yashvardhan-kukreja: Gotcha! Thanks :)

yashvardhan-kukreja (approved):
lgtm! Great work @sugarraysam , thanks 🚀

PR #35, opened 2021-12-16
Refactors

sugarraysam:
Branched off of #34 , after it is merged and this PR rebased, the diff will look much more reasonable.

## Collection of refactors for validators

- Add locks on singleton and extraction (found race conditions)
- `Validator interface` -> `ValidatorTest interface` and moved to `types.go`
- Delete the `csv_current` example because channels are to be deprecated what matters is that the `operatorName` matches the package name annotation (in next PR)
- Add a few tests, parallelize tests
- Rename a few files


yashvardhan-kukreja:
@sugarraysam Heads up: #34 is merged. Can you rebase this PR to the HEAD of master ? :)

sugarraysam:
@yashvardhan-kukreja PTAL, includes #38

sugarraysam:
Simplified error handling in test code helpers, simply `log.Fatal(...)` as anyway this is only testing code.

pkg/validators/test_001_default_channel.go
```diff
@@ -0,0 +1,28 @@
...
+	return Validate001DefaultChannel(mb)
+}
+
+func (v ValidatorTest001DefaultChannel) SucceedingCandidates() []utils.MetaBundle {
+	return testutils.DefaultSucceedingCandidates()
+}
+
+// not implemented
+func (v ValidatorTest001DefaultChannel) FailingCandidates() []utils.MetaBundle {
+	return []utils.MetaBundle{}
```
yashvardhan-kukreja, line 27: shouldn't we add the previously added failingCandidates here as well for now ?

pkg/utils/types.go
```diff
@@ -6,10 +6,18 @@ import (
 )
 
 type Validator struct {
+	Name        string
```
yashvardhan-kukreja, line 9: Why is this attribute needed? The `AllValidators` seems to track it via it's key right and at the same time, ensures uniqueness of them :)

yashvardhan-kukreja:
Great work!
Just some aforementioned questions, rest all lgtm :)

pkg/utils/types.go
sugarraysam: If you are familiar with Hadolint and ShellCheck, #40 is the first step towards a similar approach.

The idea is, there is no `failureMsg` good enough to fully explain a static check failure, so we will instead populate a wiki: https://github.com/mt-sre/addon-metadata-operator/wiki/Home, with 1 page per validator.

Upon failure, we will throw a standardized failureMsg, that points to the wiki. The `Name` attribute can be like pylint, where you have a code `AMXXXX` that doesn't mean anything, but the name is more human friendly.

```go
func (v Validator) Fail(failureMsg string) (bool, string, error) {
  msg := fmt.Sprintf("%v: (%v) %v (%v)", v.Code, v.Name, failureMsg, v.GetWikiURL)
  return false, msg, nil
}
```

WDYT?

pkg/validators/test_001_default_channel.go
sugarraysam: done

pkg/utils/types.go
yashvardhan-kukreja: Got it! Really cool stuff.

yashvardhan-kukreja (approved):
lgtm! Thanks :)

PR #37, opened 2021-12-16
Add 007_csv_install_modes validator

sugarraysam:
To be merged after #34  and #35

### DIFF
--- a/pkg/validators/am0007_csv_install_modes.go
+++ b/pkg/validators/am0007_csv_install_modes.go
@@ -1,0 +1,74 @@
+package validators
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/mt-sre/addon-metadata-operator/pkg/utils"
+	operatorsv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
+	"github.com/operator-framework/operator-registry/pkg/registry"
+)
+
+func init() {
+	Registry.Add(AM0007)
+}
+
+var AM0007 = utils.Validator{
+	Code:        "AM0007",
+	Name:        "csv_install_modes",
+	Description: "Validate installMode is supported.",
+	Runner:      validateCSVInstallModes,
+}
+
+func validateCSVInstallModes(mb utils.MetaBundle) (bool, string, error) {
+	installMode := mb.AddonMeta.InstallMode
+
+	for _, bundle := range mb.Bundles {
+		bundleName, err := utils.GetBundleNameVersion(bundle)
+		if err != nil {
+			return false, "", err
+		}
+		spec, err := extractCSVSpec(bundle)
+		if err != nil {
+			return false, "", fmt.Errorf("Can't extract CSV for %v, got %v.", bundleName, err)
+		}
+		if success, failureMsg := isInstallModeSupported(spec.InstallModes, installMode); !success {
+			return false, fmt.Sprintf("Bundle %v failed CSV validation: %v.", bundleName, failureMsg), nil
+		}
+	}
+	return true, "", nil
+}
+
+func isInstallModeSupported(installModes []operatorsv1alpha1.InstallMode, target string) (bool, string) {
+	var allSupported []string
+
+	targetSupported := false
+	for _, im := range installModes {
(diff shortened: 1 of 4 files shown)


pkg/validators/test_am0007.go
```diff
@@ -0,0 +1,29 @@
...
+}
+
+func (v TestAM0007) SucceedingCandidates() []utils.MetaBundle {
+	return testutils.DefaultSucceedingCandidates()
+}
+
+// not implemented
+func (v TestAM0007) FailingCandidates() []utils.MetaBundle {
+	return []utils.MetaBundle{}
+}
```
yashvardhan-kukreja, line 29: Let's add some failing candidates here?

pkg/validators/am0007_csv_install_modes.go
```diff
@@ -0,0 +1,72 @@
...
+		if success, failureMsg := isInstallModeSupported(spec.InstallModes, installMode); !success {
+			return false, fmt.Sprintf("Bundle %v failed CSV validation: %v.", bundleName, failureMsg), nil
+		}
+	}
+	return true, "", nil
+}
+
+func isInstallModeSupported(installModes []operatorsv1alpha1.InstallMode, target string) (bool, string) {
+	var allSupported []string
+	for _, im := range installModes {
```
yashvardhan-kukreja, line 44:
```suggestion
	var allSupported []string
	targetImSupported := false
	for _, im := range installModes {
```

pkg/validators/am0007_csv_install_modes.go
```diff
@@ -0,0 +1,72 @@
...
+		imType := string(im.Type)
+		if im.Supported {
+			allSupported = append(allSupported, imType)
+		}
+
+		if imType == target {
+			return im.Supported, ""
+		}
```
yashvardhan-kukreja, line 52:
```suggestion
		if imType == target {
			targetImSupported = im.Supported
		}
```

pkg/validators/am0007_csv_install_modes.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+	return false, fmt.Sprintf("Target installMode %v is not supported. CSV only supports these installModes %v.", target, allSupported)
```
yashvardhan-kukreja, line 54: So as to generate a relevant failureStr instead of straight away returning at `return im.Supported, ""` (Original Line: 52)

```suggestion
	targetImSupported := false
	for _, im := range installModes {
		imType := string(im.Type)
		if im.Supported {
			allSupported = append(allSupported, imType)
		}

		if imType == target {
			targetImSupported = im.Supported 
		}
	}
	failureStr := ""
	if !targetImSupported {
	    failureStr = fmt.Sprintf("Target installMode %v is not supported. CSV only supports these installModes %v.", target, allSupported)
	}
	return targetImSupported, failureStr 
```

yashvardhan-kukreja (changes requested):
@sugarraysam Just some minor changes mentioned above, rest all looks great :) Thanks!

Can you rebase this PR and resolve the conflicts as well?

pkg/validators/test_am0007.go
sugarraysam: This is tricky, as we would need to package failing bundles, and we don't have an easy way to do this right now.

pkg/validators/am0007_csv_install_modes.go
sugarraysam: Good catch, if `im.Supported` is false we would get an empty failure message. :)

pkg/validators/test_am0007.go
yashvardhan-kukreja: Ahh makes sense. I just looked into it. We'd have to 
- define a static YAML/JSON of a CSV 
- unmarshal it into an `unstructured.Unstructured` object 
- feed that object to `registry.NewBundle(.., ..., obj)`
- And finally, build a failing `[]utils.MetaBundle` out of it.

I can wrap it up in a separate PR :)
This PR looks good to me :)

yashvardhan-kukreja (approved):
lgtm

PR #38, opened 2021-12-17
fix: data races amongst fighting concurrent test-goroutines

yashvardhan-kukreja:
Signed-off-by: Yashvardhan Kukreja <ykukreja@redhat.com>

PR #39, opened 2021-12-17
MTSRE-373 | add: dms snitchNamePostFix validator

yashvardhan-kukreja:

### DIFF
--- a/pkg/validators/am0006_dms_snitchnamepostfix.go
+++ b/pkg/validators/am0006_dms_snitchnamepostfix.go
@@ -1,0 +1,30 @@
+package validators
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/mt-sre/addon-metadata-operator/pkg/utils"
+)
+
+func init() {
+	Registry.Add(AM0006)
+}
+
+var AM0006 = utils.Validator{
+	Code:        "AM0006",
+	Name:        "dms_snitchnamepostfix",
+	Description: "Ensure `deadmanssnitch.snitchNamePostFix` doesn't begin with 'hive-'",
+	Runner:      ValidateDmsSnitchNamePostFix,
+}
+
+func ValidateDmsSnitchNamePostFix(metabundle utils.MetaBundle) (bool, string, error) {
+	dmsConf := metabundle.AddonMeta.DeadmansSnitch
+	if dmsConf == nil || dmsConf.SnitchNamePostFix == nil {
+		return true, "", nil
+	}
+	if strings.HasPrefix(*dmsConf.SnitchNamePostFix, "hive-") {
+		return false, fmt.Sprintf("`deadmanssnitch.snitchNamePostFix` in addon %s found to begin with 'hive-'", metabundle.AddonMeta.ID), nil
+	}
+	return true, "", nil
+}
--- a/pkg/validators/helpers.go
+++ b/pkg/validators/helpers.go
@@ -1,0 +1,3 @@
+package validators
+
+func StringToPtr(s string) *string { return &s }
--- a/pkg/validators/test_am0006.go
+++ b/pkg/validators/test_am0006.go
@@ -1,0 +1,63 @@
+package validators
+
+import (
+	"github.com/mt-sre/addon-metadata-operator/api/v1alpha1"
+	"github.com/mt-sre/addon-metadata-operator/internal/testutils"
+	mtsrev1 "github.com/mt-sre/addon-metadata-operator/pkg/mtsre/v1"
+	"github.com/mt-sre/addon-metadata-operator/pkg/utils"
+)
+
+func init() {
+	TestRegistry.Add(TestAM0006{})
+}
+
+type TestAM0006 struct{}
+
+func (t TestAM0006) Name() string {
+	return AM0006.Name
+}
+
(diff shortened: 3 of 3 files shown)


yashvardhan-kukreja:
please merge #38 before merging this PR

internal/testutils/reference_addon.go
```diff
@@ -103,6 +109,9 @@ func (r *ReferenceAddonStage) GetImageSet(version string) (*addonsv1alpha1.Addon
 }
 
 func (r *ReferenceAddonStage) GetMetaBundle(version string) (*utils.MetaBundle, error) {
```
ashishmax31, line 111: is there a reason why we have made defined this method on a pointer receiver ?
We can get rid of the deepcopy stuff if we defined this on a normal object.

ashishmax31: We can also get rid of `metaBundleMutex` if we change the method receiver to a normal object. (`ExtractAndParseAddons` fn has its own internal mutex)

sugarraysam: The `ReferenceAddonStage` is a singleton that we initialize once by reading yaml manifests. This is to speedup testing, we load the metadata once globally from `internal/testdata/`. Than we generate the `metaBundle` object using deepcopy instead of reading yaml manifests again. Deepcopying is faster than I/O, and allows each call to `GetMetaBundle` to get a fresh, independent `metaBundle`.

sugarraysam: The `metaBundleMutex` is superseded by the global mutex on `ExtractAndParseAddons`.

@yashvardhan-kukreja please rebase this PR.

ashishmax31: @sugarraysam What I was thinking is this, 
```
meta := *r.MetaImageSet
if err := meta.PatchWithImageSet(imageSet); err != nil ...
```
But i guess this doesn't actually clone everything as `r.MetaImageSet` has some pointer attributes.
Lets stick to deepcopying

sugarraysam (dismissed):
lgtm

pkg/validators/am0006_dms_snitchnamepostfix.go
```diff
@@ -0,0 +1,30 @@
...
+var AM0006 = utils.Validator{
+	Code:        "AM0006",
+	Name:        "dms_snitchnamepostfix",
+	Description: "Ensure `deadmanssnitch.snitchNamePostFix` doesn't begin with 'hive-'",
+	Runner:      ValidateDmsSnitchNamePostFix,
+}
+
+func ValidateDmsSnitchNamePostFix(metabundle utils.MetaBundle) (bool, string, error) {
+	dmsConf := metabundle.AddonMeta.DeadmansSnitch
+	if dmsConf == nil || dmsConf.SnitchNamePostFix == nil {
```
sugarraysam, line 23: We are accepting no `dmsConf` or empty `SnitchNamePostFix`. Is this intended? I have no context on what we want from DMS, and want to enforce for our tenants. 

cc @apahim 

Are there other things we want to validate in the `DeadmansSnitch` config?

sugarraysam (commented):
If you could add some wiki docs that would be awesome: https://github.com/mt-sre/addon-metadata-operator/wiki/AM0006

pkg/validators/am0006_dms_snitchnamepostfix.go
yashvardhan-kukreja: Yep. All the fields under the `deadmanssnitch` section are templatised as a default manner. Although, we still give the option to the tenants to override the child fields like `snitchNamePostFix` in which case the above check makes sense.

> Are there other things we want to validate in the DeadmansSnitch config?
For now, none. I can think of one thing and that is that if `deadmanssnitch` field is there, then`deadmanssnitch.tags` must be there as well.

But the field level checks which will be added later should suffice for that.


https://github.com/mt-sre/addon-metadata-operator/blob/master/pkg/mtsre/v1/addons.go#L92-L93

yashvardhan-kukreja:
> If you could add some wiki docs that would be awesome: https://github.com/mt-sre/addon-metadata-operator/wiki/AM0006

Done @sugarraysam

pkg/validators/am0006_dms_snitchnamepostfix.go
sugarraysam: Interesting, actually I opened #42  to remove the field-level checks, as it seemed they were not required anymore. I felt having both `validate:"required"` tags and `// +kubebuilder:validation:Required` annotations was a duplication of the same feature.

The k8s API server will enforce all those kubebuilder annotations upon CR creation. 

WDYT?

yashvardhan-kukreja:
@sugarraysam I also checked this validator against our current addons in integration, stage and production envs. I even checked this validator against the HEAD of the branch (yet to be merged) which migrates all the addons depending on dms to start making use of `deadmanssnitch` section instead of `extraResources` section.

All the above situations work perfectly fine with the above validator, so I guess we're good to go.

yashvardhan-kukreja:
@sugarraysam I believe that after the above PR is merged and is in use, we can even get rid of https://github.com/mt-sre/managed-tenants-cli/blob/main/managedtenants/core/addons_loader/sss.py#L76-L102

sugarraysam (approved):

sugarraysam:
lgtm

PR #41, opened 2021-12-23
Test harness validator

ashishmax31:
Adds a new validator for ensuring the presence of test-harness image in the addon's metadata file.

### DIFF
--- a/pkg/validators/am0005_test_harness.go
+++ b/pkg/validators/am0005_test_harness.go
@@ -1,0 +1,70 @@
+package validators
+
+import (
+	"fmt"
+	"net/http"
+	"net/url"
+	"time"
+
+	"github.com/mt-sre/addon-metadata-operator/pkg/utils"
+	imageparser "github.com/novln/docker-parser"
+)
+
+const quayRegistryApi = "https://quay.io/v2"
+
+var AM0005 = utils.Validator{
+	Code:        "AM0005",
+	Name:        "test_harness",
+	Description: "Ensure that an addon has a valid testharness image",
+	Runner:      ValidateTestHarness,
+}
+
+func init() {
+	Registry.Add(AM0005)
+}
+
+func ValidateTestHarness(metabundle utils.MetaBundle) (bool, string, error) {
+	res, err := imageparser.Parse(metabundle.AddonMeta.TestHarness)
+	if err != nil {
+		return false, "Failed to parse testharness url", err
+	}
+	if res.Registry() != "quay.io" {
+		return false, "Testharness image is not in the quay.io registry", nil
+	}
+	retryCount := 5
+	imagePresent, err := checkImageExists(res, retryCount)
+	if err != nil {
+		return false, "Encountered an error when trying to access the remote image", err
+	}
+	if !imagePresent {
+		return false, "Test harness image doesnt exist", nil
+	}
+	return true, "", nil
+}
+
+func checkImageExists(imageUri *imageparser.Reference, retryCount int) (bool, error) {
+	apiUrl := fmt.Sprintf("%s/%s/manifests/%s", quayRegistryApi, imageUri.ShortName(), imageUri.Tag())
+	resp, err := http.Get(apiUrl)
+	if err != nil {
+		if retryCount > 0 && retryableError(err) {
+			time.Sleep(2 * time.Second)
+			return checkImageExists(imageUri, retryCount-1)
+		}
+		return false, err
+	}
+
+	if resp.StatusCode == 200 {
+		return true, nil
+	}
+	// Retry on 5xx responses
+	if retryCount > 0 && resp.StatusCode >= 500 {
+		time.Sleep(2 * time.Second)
+		return checkImageExists(imageUri, retryCount-1)
+	}
+	return false, nil
+}
+
+func retryableError(err error) bool {
+	urlErr, ok := err.(*url.Error)
+	return ok && urlErr.Timeout()
+}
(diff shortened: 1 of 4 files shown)


sugarraysam:
I cant use the review feature from Github?? Is this because of the conflicting files?

ashishmax31:
@sugarraysam thats weird, I'm able to see my changes

pkg/validators/validate_test_harness.go
```diff
@@ -0,0 +1,43 @@
...
+	if err != nil {
+		return false, "Encountered an error when trying to access the remote image", err
+	}
+	if !imagePresent {
+		return false, "Test harness image doesnt exist", nil
+	}
+	return true, "", nil
+}
+
+func checkImageExists(imageUri *imageparser.Reference) (bool, error) {
```
sugarraysam, line 33: As `quay.io` has proven quite flaky recently, do you think it's worth adding a retry mechanism for more reliability?

sugarraysam (approved):
lgtm, please rebase and follow naming convention.

Also, after picking your `AMxxxx` code, it would be nice to add a wiki page: https://github.com/mt-sre/addon-metadata-operator/wiki/AMxxxx.

View `default_channel` example: https://github.com/mt-sre/addon-metadata-operator/wiki/AM0001

ashishmax31:
@sugarraysam Have a look now. I'll add the wiki after this is merged.

pkg/validators/am0005_test_harness.go
```diff
@@ -0,0 +1,81 @@
+// ValidateAddonLabel validates whether the 'label' field under an addon.yaml follows the format 'api.openshift.com/addon-<id>'
+// TODO - remove this validator once we do field level validation
```
sugarraysam, line 2: nit: remove leftover comments

pkg/validators/am0005_test_harness.go
```diff
@@ -0,0 +1,81 @@
...
+
+func ValidateTestHarness(metabundle utils.MetaBundle) (bool, string, error) {
+	res, err := imageparser.Parse(metabundle.AddonMeta.TestHarness)
+	if err != nil {
+		return false, "Failed to parse testharness url", err
+	}
+	if res.Registry() != "quay.io" {
+		return false, "Testharness image is not in the quay.io registry", nil
+	}
+	imagePresent, err := checkImageExists(res, 5)
```
sugarraysam, line 35: nit: add a var named `retryCount` so we understand the code easily -> `checkImageExists(res, retryCount)`

pkg/validators/am0005_test_harness.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+	// Retry on 5xx responses
+	if retryableResponse(resp.StatusCode, retryCount) {
+		return checkImageExists(imageUri, retryCount-1)
+	}
+	return false, nil
+}
+
+func retryableError(err error, currentRetryCount int) bool {
+	if err != nil {
```
sugarraysam, line 66: suggested refactor:

```go
func retryableError(err error) bool {
  urlErr, ok := err.(*url.Error)
  return ok && urlErr.Timeout()
}
```

pkg/validators/am0005_test_harness.go
```diff
@@ -0,0 +1,81 @@
...
+	if !imagePresent {
+		return false, "Test harness image doesnt exist", nil
+	}
+	return true, "", nil
+}
+
+func checkImageExists(imageUri *imageparser.Reference, retryCount int) (bool, error) {
+	apiUrl := fmt.Sprintf("%s/%s/manifests/%s", quayRegistryApi, imageUri.ShortName(), imageUri.Tag())
+	resp, err := http.Get(apiUrl)
+	if err != nil {
```
sugarraysam, line 48: Suggest a refactor:
- add `time.Sleep()` -> could get fancy and use `retryCount`
- simplify logic in one line
- `retryableError(err)` has 1 arg

```go
if err != nil && retryCount > 0 && retryableRequestError(err) {
  time.Sleep(2 * time.Second)
  return checkImageExists(imageUri, retryCount-1)
}
```

pkg/validators/am0005_test_harness.go
```diff
@@ -0,0 +1,81 @@
...
+		if retryableError(err, retryCount) {
+			return checkImageExists(imageUri, retryCount-1)
+		}
+		return false, err
+	}
+	if resp.StatusCode == 200 {
+		return true, nil
+	}
+	// Retry on 5xx responses
+	if retryableResponse(resp.StatusCode, retryCount) {
```
sugarraysam, line 59: Suggested refactor, get rid of extra func:

```go
if retryCount > 0 && resp.StatusCode >= 500 {
  time.Sleep(2 * time.Second)
  checkImageExists(imageUri, retryCount-1)
}
```

sugarraysam (commented):
Left few nits, suggested refactors.

Please make sure to test against `managed-tenants` following the validators SOP before merging.

pkg/validators/am0005_test_harness.go
ashishmax31: If we simplify the logic in one line (`err != nil && retryCount > 0 && retryableRequestError(err) `) We'll have to add check `resp ` is not nil for all the lines below it which expect response to be non nil. Also, we still need to add a
branch which returns the error. 
What about this instead?
```
if err != nil {
	if retryCount > 0 && retryableError(err) {
		time.Sleep(2 * time.Second)
		return checkImageExists(imageUri, retryCount-1)
	}
	return false, err
}
```

sugarraysam (approved):
lgtm

PR #42, opened 2021-12-27
Remove go-playground validate tags.

sugarraysam:
I think it doesn't make sense to use the go-playground framework anymore.

It is simpler to write well defined validators to enforce field-level validation, than to write a function, and add a shady tag to a type description. The user experience for our tenants is better with validators as they will provide extensive documentation in their wikis.


sugarraysam:
Actually it's a good idea to keep these validator tags.

It seemed we were duplicating the kubebuilder annotations feature, but, we need to remember those are not enforced when `$ mtcli validate` is ran as a pure CLI based approach to validation.

Closing.

PR #43, opened 2021-12-27
New output formatting for ValidateCLI.

sugarraysam:
Using new fields in the Validator struct to provide better validation output.

Advertise the new wiki: https://github.com/mt-sre/addon-metadata-operator/wiki/<code>

Fix output breaking with inconsistent tabs.

Example of new output:

```bash
$ go run cmd/mtcli/main.go validate ../managed-tenants/addons/reference-addon
```
image

### DIFF
--- a/pkg/validate/print.go
+++ b/pkg/validate/print.go
@@ -1,32 +1,49 @@
 package validate
 
 import (
-	"fmt"
-	"strings"
-
+	"github.com/alexeyco/simpletable"
 	"github.com/mt-sre/addon-metadata-operator/pkg/utils"
 )
 
 var (
-	equalLines = strings.Repeat("=", 6)
-
-	success = utils.Green("Success")
-	failed  = utils.Red("Failed")
-	err     = utils.IntenselyBoldRed("Error")
+	statusSuccess = utils.Green("Success")
+	statusFailed  = utils.Red("Failed")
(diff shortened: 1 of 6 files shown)


pkg/validate/print.go
```diff
@@ -1,32 +1,49 @@
...
 
-func printFailureMessage(msg string) {
-	fmt.Printf("\r%s\t\t%s", msg, failed)
+func newErrorTableRow(v utils.Validator, err error) []*simpletable.Cell {
+	return newTableRow(v, statusError, err.Error())
 }
 
-func printErrorMessage(msg string) {
-	fmt.Printf("\r%s\t\t%s", msg, err)
+func newTableRow(v utils.Validator, status, failureMsg string) []*simpletable.Cell {
```
yashvardhan-kukreja, line 38: Any specific reason why these functions are intended to return pointer `[]*simpletable.Cell` and not value `[]simpletable.Cell`?

yashvardhan-kukreja:
apart from the above question of mine, rest all lgtm :) great stuff!

pkg/validate/print.go
sugarraysam: Because on the other side, in `validate.go`, you have `t.Header` + `t.Body.Cells` which require pointers

yashvardhan-kukreja (approved):

PR #45, opened 2021-12-31
Add ValidatorResult and refactor pkg/utils/types.go

sugarraysam:
Adding a wrapper around `ValidatorResult`.

This makes the code easier to read/maintain, and opens the door to a retry middleware (next PR):

```go
type Validator {
  // ...
  // add two fields for retry logic
  RetryCount int
  RetryDelaySeconds int
}

type Middleware func(Validator) Validator

func RetryMiddleware(v Validator, mb MetaBundle) Validator {
  runner := func(mb MetaBundle) ValidatorResult {
    retryCount := 5
    retryDelaySeconds := 2
    
    if v.RetryCount > 0 {
      retryCount = v.RetryCount
    }
    if v.RetryDelaySeconds > 0 {
      retryDelaySeconds = v.RetryDelaySeconds
    }
    
    var res ValidatorResult
    for retryCount > 0; retryCount-- {
      res = v.Runner(mb)
      if !res.IsRetryableError() {
        return res
      }
      time.Sleep(retryDelaySeconds * time.Seconds)
    }
    // retryCount == 0 and err still retryable
    return res
  }
  return v.WithRunner(runner) // helper
}
```


pkg/types/validators.go
```diff
@@ -0,0 +1,54 @@
...
+// ValidatorResult - encompasses validator result information
+type ValidatorResult struct {
+	// true if MetaBundle validation was successful
+	Success bool
+	// "" if validation is successful, else information about why it failed
+	FailureMsg string
+	// retports error that happened in the validation code
+	Error error
+	// if an error occured in the validation code, determines if it was retryable
+	RetryableError bool
```
Ajpantuso, line 45: Not to rework this too much, but are `Error` and `RetryableError` really members of the result? Or does a `ValidateFunc` want to return `(ValidatorResult, error)` with _retryable errors_ wrapped in a custom error type that can be detected with `errors.Is`? I feel the latter is more idiomatic and less prone to someone confusing when to use `Fail()`, `Error()`, and `RetryableError()`.

Ajpantuso (commented):
Looks sensible, but raised a question in review comments.

pkg/types/validators.go
sugarraysam: I personally like the single return value. It's simpler, you only have to manipulate this one result, and all cases are covered. If our goal is to abstract a ValidatorResult, and an error condition is a valid result, than why not bake it into the abstraction?

sugarraysam: And say we split the two, then what becomes the ValidatorResult in case an error occurs? `return Fail(), err` ? `return nil, err` ? Then we should use pointers. IDK. And we would have a bunch of `return Success(), nil`, `return Fail(), nil` all over the place.

Ajpantuso: Simplicity is nice, but as you add MW there may be situations where errors no longer occur during validation, but within the MW logic. You could pass those errors as part of the result, but failed/errored validation is not handled like a normal error so things will get weird. Also, the signature of your helpers could change e.g. `func Fail() (ValidatorResult, error)` and not sure pointers are necessary as long as defaults are useful (which I think they are in this case since `Success := false` makes sense as a default). I could be swayed either way though which is why this is just a comment.

Ajpantuso: Still Major version 0 so there's time to play around using the single result. As for name change maybe nouns representing result state i.e. `Success()`, `Failure(msg)`, `InError(err)`? That way `return Failure(msg)` reads distinctly different than `return InError(err)` and we don't need a negative name.

sugarraysam: I like `Failure()`, but... I don't like `InError()` lolll ...we'll get there.  :rofl:

What do you guys think? Ashish @ashishmax31 ? Yash @yashvardhan-kukreja ?

Ajpantuso: Static error types/checking definitely does add complexity to the interface, but if `ValidatorResult` is analogous to `http.Response` then it should be noted that _clients_ can also check for errors outside of the HTTP status code (See [here](https://cs.opensource.google/go/go/+/refs/tags/go1.17.5:src/net/http/client.go;drc=ab175939e28df62fee09c33c7d9e008bff6b684c;l=456)). But the original concern was around confusing devs so if the team prefers a single result object then that's the way to go.

Ajpantuso: After discussion with Yash we've decided that a single result object is sufficient to start making progress on the overall refactor permitting middleware. Any revelations regarding a more optimal design will come with further iterations (which requires this PR to be merged).

validators_test.md
```diff
@@ -12,32 +12,33 @@ Every validator test is represented by a dedicated type (struct) which implement
 
 ```go
 type Validator interface {
-	Name() string
-	Run(utils.MetaBundle) (bool, error)
-	SucceedingCandidates() []utils.MetaBundle
-	FailingCandidates() []utils.MetaBundle
+    Name() string
+    Run(MetaBundle) (bool, error)
```
Ajpantuso, line 16:
```suggestion
    Run(types.MetaBundle) (bool, error)
```

sugarraysam: good catch, fixed to -> `Run(types.MetaBundle) types.ValidatorResult`

removed the commented out code and pasted in the ticket as a comment instead -> https://issues.redhat.com/browse/MTSRE-390

Ajpantuso (approved):
lgtm

PR #49, opened 2022-01-05
[MTSRE-384] IndexExtractor for both sql-based and file-based catalog

sugarraysam:
To be merged after #45 as it relies on types being in their own package.

An interesting thing is that we need to add `export CGO_CFLAGS="-DSQLITE_ENABLE_JSON1"` as otherwise `go-sqlite3` is built without the JSON extension, and we get cryptic errors when working with sql-based catalog:

```bash
populate channels: no such function: json_object
```

### DIFF
--- a/pkg/extractor/index.go
+++ b/pkg/extractor/index.go
@@ -1,0 +1,66 @@
+package extractor
+
+import (
+	"context"
+	"sort"
+
+	"github.com/operator-framework/operator-registry/alpha/action"
+	"github.com/operator-framework/operator-registry/alpha/model"
+)
+
+// allBundlesKey - special cacheKey that means "list all bundles for all packages in the indexImage"
+const allBundlesKey = "__ALL__"
+
+type indexExtractor struct {
+	IndexExtractorCache
+}
+
+func NewIndexExtractor(cache IndexExtractorCache) IndexExtractor {
+	return &indexExtractor{cache}
+}
+
+// ListBundlesFromPackage - returns a sorted list of bundles for a given pkg
+func (i *indexExtractor) ListBundlesFromPackage(indexImage string, pkgName string) ([]string, error) {
+	return i.listBundlesWithCache(indexImage, pkgName)
+}
+
+// ListAllBundles - returns a sorted list of all bundles for all pkgs
+func (i *indexExtractor) ListAllBundles(indexImage string) ([]string, error) {
+	return i.listBundlesWithCache(indexImage, allBundlesKey)
+}
+
+// listBundles - return a list of all bundle images, sorted
+func (i *indexExtractor) listBundlesWithCache(indexImage string, cacheKey string) ([]string, error) {
+	if bundles := i.GetBundles(indexImage, cacheKey); bundles != nil {
+		return bundles, nil
+	}
+	pkgName := pkgNameFromCacheKey(cacheKey)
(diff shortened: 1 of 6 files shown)


sugarraysam:
@ashishmax31 PTAL

pkg/extractor/index.go
```diff
@@ -0,0 +1,36 @@
...
+	return indexExtractor{cache}
+}
+
+// ListBundles - return a list of all bundle images, sorted
+func (i indexExtractor) ListBundles(indexImage string, packageName string) ([]string, error) {
+	if bundles := i.GetBundles(indexImage, packageName); bundles != nil {
+		return bundles, nil
+	}
+	res := make([]string, 0)
+	lb := action.ListBundles{IndexReference: indexImage, PackageName: packageName}
```
ashishmax31, line 25: Just wondering how we'd know the package names of all dependent operators in an indeximage.

ashishmax31:
@sugarraysam Looks good. Just had a small concern around passing packageNames of operators.

pkg/extractor/index.go
sugarraysam: So if we leave `packageName` to "", it lists bundles for all packages. So we can always add this functionality. I did not for this PR as I was wondering in what cases do we need to know the bundles from dependent operators? (e.g.: prometheus) Do we have a need/use case for it?

My understanding is that we only need to list bundles associated with the `operatorName` field from AddonMetadata (which maps to the `packageName`), so we can cross validate metadata and bundles.

pkg/extractor/index.go
```diff
@@ -0,0 +1,36 @@
...
+type indexExtractor struct {
+	types.IndexExtractorCache
+}
+
+func NewIndexExtractor(cache types.IndexExtractorCache) types.IndexExtractor {
+	return indexExtractor{cache}
+}
+
+// ListBundles - return a list of all bundle images, sorted
+func (i indexExtractor) ListBundles(indexImage string, packageName string) ([]string, error) {
```
sugarraysam, line 20: @ashishmax31 @yashvardhan-kukreja  Should this be a pointer receiver? As it modifies the `indexExtractor` object, populating it's cache?

sugarraysam: I went ahead and made it a pointer receiver, as it is modifying the state of the internal cache structure.

ashishmax31: @sugarraysam I don't think its required since `map` is a reference object by default.

pkg/extractor/index.go
ashishmax31: >Do we have a need/use case for it?
`mtcli list-bundles` should list all the operators in the image right?

pkg/extractor/index.go
sugarraysam: Digging into go FAQ, even for a `map`, which is a reference type object, its best practice to make the method receiver a pointer if it modifies internal state: https://go.dev/doc/faq#methods_on_values_or_pointers

> First, and most important, does the method need to modify the receiver? If it does, the receiver must be a pointer.

pkg/extractor/index.go
sugarraysam: That's right, ok let me make it obvious `ListBundles` vs `ListAllBundles`

Ajpantuso (approved):
lgtm

PR #51, opened 2022-01-10
MTSRE-388 | add support for validator assets and add candidates for 0007 validator

yashvardhan-kukreja:

It's tough to write succeeding/failing candidates for validators which involving fiddling with static assets like CSV and what not. Hence, having a `internal/testdata/assets/am{validator-id}` dedicated to such scenarios to store any number of static assets if need be for the testing of a validator.

For example: This PR itself depicts how 0007 validator's candidates make use of csv.yaml manifest provided under `internal/testdata/assets/am0007/` for its candidates.


pkg/validators/helpers.go
```diff
@@ -1,3 +1,35 @@
...
+
+	yamlAbsPath, err := filepath.Abs(yamlPath)
+	if err != nil {
+		return unstructured.Unstructured{}, err
+	}
+	yamlBytes, err := ioutil.ReadFile(yamlAbsPath)
+	if err != nil {
+		return unstructured.Unstructured{}, err
+	}
+	var csvObj unstructured.Unstructured
```
sugarraysam, line 30: Can we load other k8s objects with `unstructured.Unstructured`? The name `csvObj` suggests it would only be a csv.

pkg/validators/test_am0007.go
```diff
@@ -20,10 +22,63 @@ func (v TestAM0007) Run(mb utils.MetaBundle) (bool, string, error) {
...
+		panic(err)
+	}
+
+	return []utils.MetaBundle{
+		{
+			AddonMeta: &v1alpha1.AddonMetadataSpec{
+				InstallMode: "something",
+			},
+			Bundles: []registry.Bundle{
+				*registry.NewBundle("random-bundle", &registry.Annotations{}, &csvObj),
```
sugarraysam, line 64: Great idea. Thinking we could have a `testutils.NewBundle(<path_to_csv>)` helper, given a path, and it can `log.Fatal` like `testutils.DefaultSucceedingCandidates` on errors. WDYT?

pkg/validators/helpers.go
```diff
@@ -1,3 +1,35 @@
...
+	"path/filepath"
+	"strings"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/util/yaml"
+)
+
 func StringToPtr(s string) *string { return &s }
+
+func yamlToDynamicObj(yamlPath string) (unstructured.Unstructured, error) {
```
sugarraysam, line 15: nit: Would move to `utils` package as it's not really part of validators.

sugarraysam (approved):
Great stuff. Thank you for this! Left some nits but lgtm.

pkg/validators/helpers.go
yashvardhan-kukreja: Ohh yes, forgot to rename this haha. Will do :)

pkg/validators/assets/am0007/csv.yaml
```diff
@@ -0,0 +1,92 @@
+apiVersion: operators.coreos.com/v1alpha1
```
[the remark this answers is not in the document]
yashvardhan-kukreja: Sure, makes sense :)

pkg/validators/test_am0007.go
yashvardhan-kukreja: Yep, sounds good. Done :)

pkg/validators/helpers.go
yashvardhan-kukreja: Gotten rid of this entire file xD

yashvardhan-kukreja:
@sugarraysam addressed all your suggestions. You can checkout the `testutils.NewBundle(....)` helper and if it looks good to you, then please go ahead with merging this PR. Thanks :)

internal/testutils/testutils.go
```diff
@@ -45,3 +75,15 @@ func DefaultSucceedingCandidates() []types.MetaBundle {
 
 	return res
 }
+
+func NewBundle(name string, yamlFilePaths ...string) registry.Bundle {
```
sugarraysam, line 79: Nice !

One more thing, are we going to load bundles from assets located in a different directory? Thinking we could add `path.Join(BundleAssetsDir, "amXXXX/<filename>.yaml")` inside `NewBundle`

and maybe provide  a comment

```go
// NewBundle - create bundle from assets in `internal/testdata/assets/`
```

yashvardhan-kukreja: Good stuff! Let me fix this :)

yashvardhan-kukreja: > path.Join(BundleAssetsDir, path) inside NewBundle, where path is relative (e.g.: "am0007/csv.yaml")

Wouldn't this a bit confusing ( kinda over-abstracted) for the validator developer?
Because, I guess that from the eyes and perspective of the validator developer, while coding out the `test_am0007.go` they would be working inside the `pkg/validators/` directory in that specific moment. And from their eyes, the assets files would sit inside "../../internal/testdata/assets/am0007/" and that's the first thing which will pop in their mind to feed as the yamlFilePaths.
But just "am0007/*" would feel weird unless they exactly know that behind the scenes, the `NewBundle` function is prefixing their input path with "../../internal/testdata/assets". 

Sounds a lot for the validator developer (considering myself to be in the shoes of one) but wdyt :)

sugarraysam: that sounds perfectly reasonable. Lets merge it ;)

PR #52, opened 2022-01-10
[MTSRE-356] Middlewares and RetryMiddleware

sugarraysam:
Implementing the middleware pattern for validators.

Includes testing and a `RetryMiddleware`. Other interesting middlewares to write:
- `PanicMiddleware`: recover from panics
- `PrintMiddleware`: register as last middleware, prints results
- `ProfilingMiddleware`: measure exec time per validator
- `PrometheusMiddleware`: when ran inside a controller, register per validator metrics

Splitting `CLIMiddlewares` from `OperatorMiddlewares`, as we will want different functionalities, based on if validation is running as part of long-running process (e.g.: register prometheus metrics).


pkg/types/validators.go
```diff
@@ -12,6 +12,23 @@ type Validator struct {
...
+// the original Runner is lost, and it causes infinite loop scenarios (e.g.: RetryMiddleware)
+func (v Validator) WithRunner(fn ValidateFunc) Validator {
+	return Validator{
+		Name:              v.Name,
+		Code:              v.Code,
+		Description:       v.Description,
+		Runner:            fn,
+		RetryCount:        v.RetryCount,
+		RetryDelaySeconds: v.RetryDelaySeconds,
+	}
```
yashvardhan-kukreja, line 31:
```suggestion
	v.Runner = fn
	return v
```

yashvardhan-kukreja (approved):
Just the above tiny nit, rest all lgtm @sugarraysam!

pkg/types/validators.go
sugarraysam: So actually that was my first implementation, but that causes an infinite loop after applying the RetryMiddleware, because the address of the original `v.Runner` is overriden. Learned this the hard way as I debugged and traced code /w delve for more than an hour lol

yashvardhan-kukreja: Yeah, you got that right. From my observations while I did the review, the issue is because of the pointer receiver updating the original object's runner itself at [this place](https://github.com/mt-sre/addon-metadata-operator/blob/e441ac017bd2d7b666016588e2b3a03e8d0b275f/pkg/validators/middlewares.go#L34) and then, this being recursively (and infinitely) being called [here](https://github.com/mt-sre/addon-metadata-operator/blob/e441ac017bd2d7b666016588e2b3a03e8d0b275f/pkg/validators/middlewares.go#L23) (internally, it would get on an infinite loop of nesting/chaining the calling of RetryMiddleware() inside RetryMiddleware() inside RetryMiddleware() .......... as the validator.Runner())

And whenever, validator runner is called by validate.go, this infinite chain of nested functions get called leading to the infinite loop scenario you're talking about.

But making use of value receiver helps with tackling the situation by leaving the original runner as it is and just returning a deep copy of the same validator but with Retrymiddleware on top of (as the middleware of) original runner , [here](https://github.com/mt-sre/addon-metadata-operator/blob/e441ac017bd2d7b666016588e2b3a03e8d0b275f/pkg/validators/middlewares.go#L34)


PS: it gets dangerous to work with closures fiddling with references to global vars 🥶

PR #53, opened 2022-01-11
Registry does not need to be concurrent safe.

sugarraysam:
~~Following a thread in addon-operator: https://github.com/openshift/addon-operator/pull/125#discussion_r781984425~~

~~We should pass mutex by value.~~

Edit: no need for registry to be concurrent safe.

### DIFF
--- a/pkg/validators/registry.go
+++ b/pkg/validators/registry.go
@@ -2,46 +2,38 @@ package validators
 
 import (
 	"log"
-	"sync"
 
 	"github.com/mt-sre/addon-metadata-operator/pkg/types"
 )
 
 // Registry - holds all registered Validators
-var Registry = NewDefaultRegistry()
+var Registry = NewValidatorsRegistry()
 
-func NewDefaultRegistry() defaultRegistry {
-	return defaultRegistry{
-		&sync.Mutex{},
-		make(map[string]types.Validator),
-	}
+func NewValidatorsRegistry() defaultRegistry {
+	return defaultRegistry{Data: make(map[string]types.Validator)}
 }
 
 type defaultRegistry struct {
-	*sync.Mutex
 	Data map[string]types.Validator
 }
 
 // Add - update the registry in a thread-safe way. Called in init() functions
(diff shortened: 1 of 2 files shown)


erdii:
I only ever see the mux beinng `RLock`ed, but not (write) `Lock`ed. Is this intentional?

sugarraysam:
There is an `Add` function that is not shown in the diff: https://github.com/mt-sre/addon-metadata-operator/blob/b8569e79029d7cf8f203a6d2c48c8818113f465c/pkg/validators/registry.go#L23

erdii:
Got it! :see_no_evil: Thanks :)

erdii:
(soft lgtm from my side then! - I have not worked on this codebase yet so I think someone who already did should post the final lgtm)

sugarraysam:
@erdii Totally makes sense. PTAL just added some tests. Refactored the `log.Fatalf()` to `log.Panicf()` for easier testing. Choosing to panic on purpose so we don't have to check errors in the `init()` function, because we would just panic:

```go
# am0001_default_channel.go
func init() {
    Registry.Add(AM0001)
}

# if we did not panic -- no way to recover
func init() {
    if err := Registry.Add(AM0001); err != nil {
        // we are checkmate here, there is really no way to recover, and so it becomes silly to check errors in every new validator
        log.Fatal(err)
    }
}
```

ashishmax31:
Looks good.
I'm just wondering if this mutex is in fact necessary. The validators are all registered using init fns in their respective files, so a normal map should suffice right? I'm not sure where the concurrent writes are happening.

sugarraysam:
@ashishmax31 I was under the impression that `init()` functions ran concurrently within a package. But after digging into it some more, it seems they run sequentially, according to lexical order of files in which they are found (within a single package): https://yourbasic.org/golang/package-init-function-main-execution-order/

Unless we have another use case, we could drop the mutex all together.

Let me refactor the code.

ashishmax31 (approved):

PR #54, opened 2022-01-12
unsilence logger in ExtractBundlesFromImage

erdii:
Signed-off-by: Josh Gwosdz <jgwosdz@redhat.com>

This will make downstream errors in `ExtractBundlesFromImage` visible.

### DIFF
--- a/pkg/utils/default_index_image_extractor.go
+++ b/pkg/utils/default_index_image_extractor.go
@@ -1,7 +1,6 @@
 package utils
 
 import (
-	"io/ioutil"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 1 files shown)


sugarraysam (approved):
lgtm

sugarraysam:
@erdii can you please rebase?

erdii:
done! sorry that it took so long!

erdii:
Looking at the output, this change may not be such a good idea at all.
Maybe we should buffer log messages from extracting the bundles and only print them in case of an error?

sugarraysam:
Sure thing. FWIW, the indexExtractor and bundleExtractor are being rewritten to support in-memory caching, sql/file-based catalogs and extract annotations.

sugarraysam:
After #68 is merged, this will be obsolete. Lets close it.

PR #55, opened 2022-01-12
WIP - [MTSRE-394] new validator for k8s resource and field names

Ajpantuso:
# AM0010

## Description

AM0010 validates any k8s resource or field names which are included with addon metadata to ensure they conform to k8s naming conventions (DNS1123 label, DNS1123 subdomain, QualifiedName, etc...).

## Checklist

- [ ] Wiki page exists: https://github.com/mt-sre/addon-metadata-operator/wiki/AMxxx
- [ ] Followed the SOP: https://gitlab.cee.redhat.com/service/managed-tenants-sops/-/blob/main/MT-SRE/sops/AMO-validators.md
- [ ] Tested against production addons in `managed-tenants/addons/`


Ajpantuso:
This PR cannot be merged until `TestAM0010` method signatures are updated for interface changes introduced with #58.

Ajpantuso:
Conflicts resolved.

pkg/validators/am0010_k8s_resource_names.go
```diff
@@ -0,0 +1,234 @@
...
+
+func joinFailureMsgs(msgs ...failureMsg) failureMsg {
+	msgStrings := make([]string, 0, len(msgs))
+
+	for _, m := range msgs {
+		msgStrings = append(msgStrings, string(m))
+	}
+
+	return failureMsg(strings.Join(msgStrings, ", "))
+}
```
sugarraysam, line 45: This could be moved to `pkg/validators/result.go` or `pkg/validators/helpers.go`

pkg/validators/am0010_k8s_resource_names.go
```diff
@@ -0,0 +1,234 @@
...
+func isValidAnnotationName(aName string) failureMsg {
+	if valid, failureReasons := reasonsToResult(utilvalidation.IsQualifiedName(strings.ToLower(aName))); !valid {
+		return prefixedFailureMsg(
+			fmt.Sprintf("\"%s\" is not a valid kubernetes annotation name", aName),
+			failureMsg(failureReasons),
+		)
+	}
+
+	return emptyMsg
+}
```
sugarraysam, line 191: nit: Is same as `isValidLabelName`, except the message and `strings.ToLower`. Is this intended?

pkg/validators/am0010_k8s_resource_names.go
```diff
@@ -0,0 +1,234 @@
...
+	msgs := make([]failureMsg, 0, len(nsNames))
+
+	for _, name := range nsNames {
+		if msg := isValidNamespaceName(name); !msg.IsEmpty() {
+			msgs = append(msgs, msg)
+		}
+	}
+
+	return msgs
+}
```
sugarraysam, line 157: nit: refactor

`areValidNamespaceNames` + `areValidAnnotationNames` + `areValidLabelNames` have an identical implementation.

Maybe call a wrapper that takes a callback:

```go
func areValidNames(names ...string, callback func(string) failureMsg) []failureMsg {
    msgs := make([]failureMsg, 0, len(names))
    for _, name := rage names {
        if msg := callback(name); !msg.IsEmpty() {
            msgs := append(msgs, msg)
        }
    }
    return msgs
}

# then each `are...` funcs look like
func areValidAnnotationNames(aNames ...string) []failureMsg {
    return areValidNames(aNames, isValidAnnotationName)
}
```

sugarraysam (commented):
Excellent unit tests, very thorough. Suggested some nits/refactoring and maybe moving the general helpers in an external file.

pkg/validators/am0010_k8s_resource_names.go
Ajpantuso: Yes, I wanted to leave this here for now because I'd like to look at more extensive changes to `ValidatorResult` in a future PR. Basically I hate the check-fix loop that happens when checkers don't report every failure the first time which is why I have this aggregation logic to join failures msgs. However, I'd like to update `ValidatorResult.FailureMsg` from a `string` to a `[]string` and then apply some of the helpers behind the scenes. A little too much for this PR though IMO.

pkg/validators/am0010_k8s_resource_names.go
Ajpantuso: Yes, the `strings.ToLower` part is the only difference as annotations allow mixed-case prefixes. As for refactoring I could write a wrapper around `IsQualifiedName` that accepts a resource type, but it's kind of trivial. I can play around with it though.

pkg/validators/am0010_k8s_resource_names.go
Ajpantuso: Hmm, yeah I guess I just stopped refactoring at some point lol. I'll give it a shot

Ajpantuso:
Applied some refactoring per review suggestions.

pkg/validators/am0010_k8s_resource_names.go
```diff
@@ -0,0 +1,252 @@
...
+	for _, name := range names {
+		if msg := validationFunc(name); !msg.IsEmpty() {
+			msgs = append(msgs, msg)
+		}
+	}
+
+	return msgs
+}
+
+func k8sNameToValidationFunc(nameType k8sName) func(string) failureMsg {
```
sugarraysam, line 181: lgtm

Thanks for refactoring. I would say this is overkill for my personal taste, and adds extra cognitive complexity where you have to jump around some more to understand the code, but I don't have strong feelings about this.

Good job on your first PR! Merge it when you are ready. :rocket:

sugarraysam (approved):

pkg/validators/am0010_k8s_resource_names.go
Ajpantuso: It should settle out once helpers are repacked, thanks.

PR #56, opened 2022-01-12
[MTSRE-397] adding bypass for tagless images required during on-boarding

Ajpantuso:
Metadata will include a placeholder value for indexImage during on-boarding which is unpullable. This PR includes a guard to bypass image pull and bundle loading in this scenario.

### DIFF
--- a/pkg/utils/bundle_utils.go
+++ b/pkg/utils/bundle_utils.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"io/ioutil"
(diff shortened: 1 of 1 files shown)


Ajpantuso:
This could also be updated to guard against a sentinel value instead of a tagless image name (as long as SOP's are updated to reflect the sentinel).

sugarraysam:
lgtm

This works as a quick fix while we find a permanent solution.

sugarraysam (approved):

PR #57, opened 2022-01-12
Revert "MTSRE-388 | add support for validator assets and add candidates for 0007 validator"

sugarraysam:
Reverts mt-sre/addon-metadata-operator#51

The `testutils.NewBundle(..)` function `log.Fatal(...)` because files don't exist after we build the binary.

```bash
2022/01/12 14:24:35 could not generate bundle: open /home/sugar/opt/internal/testdata/assets/am0007/csv.yaml: no such file or directory
exit status 1
exit 1
```

Reverting this MR as we need a better solution for our validator testing framework. Issues observed:
- Does not use a testing package `package validators`
- SucceedingCandidates + FailingCandidates don't have a concept of error in the interface


Ajpantuso (approved):
lgtm

sugarraysam:
Actually, by the time CI was done @Ajpantuso and I worked on a fix: #58

PR #58, opened 2022-01-12
[MTSRE-399] Review validators testing framework.

sugarraysam:
- Moves validators testing to external test package.
- Add error handling to ValidatorTest interface
- Removed dangerous log.Fatal calls

### DIFF
--- a/internal/testutils/reference_addon.go
+++ b/internal/testutils/reference_addon.go
@@ -3,7 +3,6 @@ package testutils
 import (
 	"fmt"
 	"io/ioutil"
-	"log"
 	"path"
 	"sync"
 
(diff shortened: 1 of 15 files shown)


Ajpantuso (approved):
lgtm

internal/testutils/reference_addon.go
```diff
@@ -35,28 +34,32 @@ var (
...
-		if err != nil {
-			log.Fatalf("Could not load imageSet metadata for reference-addon, got %v.", err)
-		}
-		instance.MetaIndexImage = metaIndexImage
-		instance.MetaImageSet = metaImageSet
-	})
-	return instance
+func GetReferenceAddonStage() (*singleton, error) {
+	lock.Lock()
+	defer lock.Unlock()
```
yashvardhan-kukreja, line 46: Curious about the shift from once.On to Locking/Unlocking everytime :)

pkg/validators/am0007_test.go
```diff
@@ -14,19 +15,25 @@ func init() {
...
 }
 
 func (v TestAM0007) Run(mb types.MetaBundle) types.ValidatorResult {
-	return AM0007.Runner(mb)
+	return validators.AM0007.Runner(mb)
 }
 
-var (
-	testBundle = testutils.NewBundle("random-bundle", "../../internal/testdata/assets/am0007/csv.yaml")
-)
```
yashvardhan-kukreja, line 26: Any reason behind shifting from calling this function once to calling it twice individually under SucceedingCandidates() and FailingCandidates()? :)

internal/testutils/reference_addon.go
Ajpantuso: The goal was to improve error handling which singletons make difficult since the `error` value needs to cross scopes as well as the instance (and we don't want to globally declare an error var). Ideally this would just become a closure, but the use of `sync.Once` implied a potential race. If there is no race however we should just get rid of the mutex altogether.

pkg/validators/am0007_test.go
Ajpantuso: Same issue regarding error handling here. Declaring this globally within the package makes error handling difficult and leads to using unnecessary panics. Calling this within function scope allows for errors to be passed gracefully back to the calling package.

internal/testutils/reference_addon.go
yashvardhan-kukreja: > but the use of sync.Once implied a potential race

I am sorry I don't get this. It's inherently meant to achieve concurrency-safe singleton pattern right? Can you exemplify any potentially race-y scenario with sync.Once?

Also, if we want to implement our own concurrency safe singleton access without unnecessary acquisition of locks, I'd suggest refactoring this function to the following one:

```go
func GetReferenceAddonStage() (*singleton, error) {
        if instance == nil {
             // only hold the lock when singleton hasn't been initialized yet
	    lock.Lock()
	    defer lock.Unlock()
            // checking `instance == nil` again in case two goroutines enter this scope and only the first one of them gets to initialize singleton and the rest of them get the opportunity to jump out without re-initializing the singleton
            if instance == nil {
            	instance = &singleton{Env: "stage"}
	        metaIndexImage, err := instance.GetMetadata(false)
	        if err != nil {
	            return nil, fmt.Errorf("Could not load indexImage metadata for reference-addon, got %v.", err)
	        }
	        metaImageSet, err := instance.GetMetadata(true)
	        if err != nil {
	            return nil, fmt.Errorf("Could not load imageSet metadata for reference-addon, got %v.", err)
	        }
	        instance.MetaIndexImage = metaIndexImage
	        instance.MetaImageSet = metaImageSet
            }
        }
	return instance, nil
}
```

yashvardhan-kukreja: > If there is no race however we should just get rid of the mutex altogether.

Unfortunately, this was happening while executing tests atleast because tests ran concurrently (due to `t.Parallel()`) and that led to data races.

Ajpantuso: >I am sorry I don't get this. It's inherently meant to achieve concurrency-safe singleton pattern right? Can you exemplify any potentially race-y scenario with sync.Once?

The point here was that since you used `sync.Once` in the first place we could not safely get rid of mutex altogether since as you explained there are races introduced by `t.Parallel()`. Under ideal circumstances we would remove the package level state as a whole.

PR #61, opened 2022-01-14
[MTSRE-403] - New validator AM0011 sku validation

Ajpantuso:
# AM0011

## Description

This validator will check that the provided `ocmQuotaName` matches an existing SKU rule in OCM for addons whose `ocmQuotaCost` is greater than `0`.

## Checklist

- [x] Wiki page exists: https://github.com/mt-sre/addon-metadata-operator/wiki/AM0011
- [x] Tested against production addons in `managed-tenants/addons/`


internal/testutils/ocm_mock.go
```diff
@@ -0,0 +1,61 @@
+package testutils
+
+import (
+	"context"
+
+	amv1 "github.com/openshift-online/ocm-sdk-go/accountsmgmt/v1"
+)
+
+// NewMockOCMClient returns a *MockOCMClient configured with the supplied options.
+func NewMockOCMClient(opts ...MockOCMClientOption) *MockOCMClient {
```
sugarraysam, line 10: If the type satisfies the interface, would we not want to return `types.OCMClient` instead?

internal/testutils/ocm_mock_test.go
```diff
@@ -0,0 +1,17 @@
+package testutils
+
+import (
+	"testing"
+
+	"github.com/mt-sre/addon-metadata-operator/pkg/validators"
+	"github.com/stretchr/testify/require"
+)
+
+func TestMockOCMClientInterface(t *testing.T) {
```
sugarraysam, line 10: If we return `types.OCMCLient` in constructor this is not required per say

pkg/utils/default_ocm_client.go
```diff
@@ -0,0 +1,146 @@
+package utils
```
sugarraysam, line 1: Is it worth to move everything ocm related to it's own package ? We will work closely with ocm when we start writing controllers, I am sure this code will be reused alot and built upon.

pkg/utils/default_ocm_client.go
```diff
@@ -0,0 +1,146 @@
...
+	code := int(e)
+
+	return code >= 500 && code < 600
+}
+
+// NewDefaultOCMClient takes a variadic slice of options to configure a default
+// OCM client and applies defaults if no appropriate option is given. An error
+// may be returned if an unusable OCM token is provided or a connection cannot
+// be initialized otherwise the default client is returned.
+func NewDefaultOCMClient(opts ...DefaultOCMClientOption) (*DefaultOCMClient, error) {
```
sugarraysam, line 35: Again, return `types.OCMClient` ?

pkg/utils/default_ocm_client.go
```diff
@@ -0,0 +1,146 @@
...
+
+	for _, opt := range opts {
+		client.Option(opt)
+	}
+
+	if client.apiURL == "" {
+		client.apiURL = apiURLStage
+	}
+
+	if client.token == "" {
```
sugarraysam, line 49: I would not keep the `token` in memory. Would use a throw-away var, and not save it in the struct. Could cause memory leaks. So basically remove L#49-54 here.

pkg/utils/default_ocm_client.go
```diff
@@ -0,0 +1,146 @@
...
+	conn   *sdk.Connection
+	token  string
+}
+
+func (c *DefaultOCMClient) initConnection() error {
+	var err error
+
+	c.conn, err = sdk.NewConnectionBuilder().
+		URL(c.apiURL).
+		Tokens(c.token).
```
sugarraysam, line 89: Here, add the `retrieveOCMTokenFromEnv()` logic, in throw-away temp func vars.

pkg/utils/default_ocm_client.go
```diff
@@ -0,0 +1,146 @@
...
+
+	return token, nil
+}
+
+// DefaultOCMClient implements the 'types.OCMClient' interface and
+// exposes methods by which validators can communicate with OCM.
+type DefaultOCMClient struct {
+	apiURL string
+	conn   *sdk.Connection
+	token  string
```
sugarraysam, line 81: Memory leak?

pkg/utils/default_ocm_client.go
```diff
@@ -0,0 +1,146 @@
...
+	}
+}
+
+// DefaultOCMClientAccessToken updates the JWT token used to connect to OCM.
+// This option only takes effect if applied prior to connection initialization.
+func DefaultOCMClientAccessToken(token string) DefaultOCMClientOption {
+	return func(c *DefaultOCMClient) {
+		c.token = token
+	}
+}
```
sugarraysam, line 146: Do we have a use case for not reading the token from env vars?

pkg/utils/default_ocm_client_test.go
```diff
@@ -0,0 +1,73 @@
...
+import (
+	"testing"
+
+	"github.com/mt-sre/addon-metadata-operator/pkg/utils"
+	"github.com/mt-sre/addon-metadata-operator/pkg/validators"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestDefaultOCMClientInterface(t *testing.T) {
```
sugarraysam, line 12: Not required per say if constructor returns `types.OCMClient`

pkg/utils/default_ocm_client_test.go
```diff
@@ -0,0 +1,73 @@
+package utils_test
```
sugarraysam, line 1: Move to `ocm_test` package ?

pkg/validators/am0011_sku_validation.go
```diff
@@ -0,0 +1,64 @@
...
+	Registry.Add(AM0011)
+}
+
+func ValidateOCMSKUExists(mb types.MetaBundle) types.ValidatorResult {
+	client, err := utils.NewDefaultOCMClient()
+	if err != nil {
+		return Error(err)
+	}
+
+	validateFunc := GenerateOCMSKUValidator(client)
```
sugarraysam, line 28: Why use a func generator? Seems the only effect is to have code indented 8 spaces because of the `return func() { <definition> }`

pkg/utils/default_ocm_client_test.go
```diff
@@ -0,0 +1,73 @@
...
+func TestDefaultOCMClientInterface(t *testing.T) {
+	t.Parallel()
+
+	require.Implements(t,
+		new(validators.OCMClient), new(utils.DefaultOCMClient),
+		"utils.DefaultOCMClient must implement the types.OCMClient interface",
+	)
+}
+
+func TestOCMResponseErrorInterface(t *testing.T) {
```
sugarraysam, line 21: Same, should `OCMResponseError` return `types.OCMError` ?

pkg/validators/am0011_test.go
```diff
@@ -0,0 +1,61 @@
...
+
+func (val TestAM0011) Run(mb types.MetaBundle) types.ValidatorResult {
+	client := testutils.NewMockOCMClient(
+		testutils.MockOCMClientValidQuotaNames(
+			"addon-reference-addon",
+			"addon-successful-candidate",
+		),
+	)
+
+	validateFunc := validators.GenerateOCMSKUValidator(client)
```
sugarraysam, line 28: Clever way to mock client :heart:

pkg/validators/am0011_sku_validation.go
sugarraysam: Edit: saw it's to mock the client in testing. Clever ! :heart:

pkg/validators/ocm_client.go
```diff
@@ -0,0 +1,39 @@
+package validators
```
sugarraysam, line 1: Any reason this is not in `types` package? We risk circular dependencies as `package validators` is imported and imports stuff, where `types` is import-only package.

pkg/validators/ocm_client.go
```diff
@@ -0,0 +1,39 @@
...
+
+import (
+	"context"
+
+	amv1 "github.com/openshift-online/ocm-sdk-go/accountsmgmt/v1"
+)
+
+// OCMError abstracts behavior required for validators to identify the underlying
+// causes of OCM related errors.
+type OCMError interface {
```
sugarraysam, line 11: Will we have various implementation of `OCMError`? Seems to me we need a type + method receiver.

pkg/validators/ocm_client.go
```diff
@@ -0,0 +1,39 @@
...
+func IsOCMServerSideError(err error) bool {
+	ocmErr, ok := err.(OCMError)
+
+	return ok && ocmErr.ServerSide()
+}
+
+// OCMClient abstracts behavior required for validators which request data
+// from OCM to be implemented by OCM API clients.
+type OCMClient interface {
+	SKURuleGetter
```
sugarraysam, line 28: Why the extra interface separation? What else will implement the SKURuleGetter besides OCMClient?

sugarraysam (commented):
Left a zillion questions :sweat_smile:  sorry. But good work. This is awesome, very clean. Mostly asking about interfaces, package organization etc.

Ajpantuso:
> Left a zillion questions sweat_smile sorry. But good work. This is awesome, very clean. Mostly asking about interfaces, package organization etc.

No worries these are critical points to discuss moving forward:

> Where should types/packages be declared?

- Interfaces should be declared by the packages which depend on them. They describe the necessary behavior required (contract) from a dependency. The most obvious example being `fmt.Stringer`. That interface is exposed so users of the `fmt` public functions can implement `String()` on their types so that the `fmt` functions know what to print when calling `<type>.String()`.

- Types implementing interfaces can be declared anywhere and ideally are not tightly coupled to a particular package.

The consequences of this are that a dedicated `types` package doesn't work well. You don't really want to expose the types themselves unless they implement particular behavior other packages use and even then without an interface it creates issues (see [logrus](https://github.com/sirupsen/logrus/blob/85981c045988a3ebe607e2f5b8a6499d86e0ddac/logrus.go#L121-L136) for an example of having to hack an interface because another package doesn't provide one). Ideally we just keep types to the single package that uses them with the exception of all the kubernetes api stuff which kubernetes provides tools to work with.

> How big should interfaces be and should they be returned by functions?

- Interfaces should be only large enough to provide the minimum behavior required to provide necessary features. It's why we accept `io.Writer` when we write, `io.Reader` when we read, and `io.ReadWriter` when we do both. That opens your functions/methods up to accept more implementations that may not be able to support all the behavior of `io.ReadWriter`. For the particular case of `OCMClient` using smaller interfaces to describe what particular validators require makes ad-hoc mocks infinitely easier like why would a mock need a `Close` method?

- Should they be returned by functions/methods. Only if the function/method could return any implementation of the interface like a factory or for things like `context.Context` which is always an interface. In our codebase you'd probably only need interfaces returned when using closures like our middleware does.

References:
Read the following sections of this [blog](https://dave.cheney.net/2016/08/20/solid-go-design)
Liskov Substitution Principle 
Interface Segregation Principle

> Why use an interface to describe errors?

See [this](https://dave.cheney.net/tag/interfaces). It just makes life easier when you need to distinguish between different types of errors like server side or client side HTTP status codes.

> Where should the default OCM Client be packaged?

Doesn't really matter because the controllers should define their own interfaces for what they expect from an `OCMClient` independent of those defined by `validators`. Though yeah `utils` is a weird place for it in the long term.

Hope that hits most of your questions.

pkg/utils/default_ocm_client.go
Ajpantuso: I'll address the token related question here:

Yep can totally nuke this from the struct. We could play around with refresh tokens if we need to, but it's not needed now.

> We will ever get OCM tokens from outside the env?

No clue, but I can remove the option and add some way to make it easy to inject the dependency for mocking.

pkg/utils/default_ocm_client.go
sugarraysam: we can revisit this later

sugarraysam:
Thanks for the comprehensive answer. I was never a big fan of centralized `types` package, only moved it from `utils` as it was better named.

Ajpantuso:
> Thanks for the comprehensive answer. I was never a big fan of centralized `types` package, only moved it from `utils` as it was better named. Probably more refactors to be done to fix that.
> 
> We can revisit moving ocm to it's package later.
> 
> Please undraft and tick the last checkbox items. Otherwise LGTM from my side.
> 
> Edit: cant approve until you undraft

I'll ready_for_review this, but put a hold for now.

sugarraysam (approved):

PR #62, opened 2022-01-14
Extending AM0010 to handle pagerDuty and bundleParameter fields with refactor

Ajpantuso:
Several items are addressed in this PR as a follow up to #55:
- Additional k8s resource names were identified in the metadata (`pagerDuty.Secret`, `pagerDuty.SecretNameSpace` and `bundleParameters.AddonParamsSecretname`) and have been added to the scope of AM0010.
- Private helpers of AM0010 have been refactored and repackaged into more appropriate locations.
- The `validate` subcommand has also been enabled to write multiple failures for a single `ValidatorResult`.

### DIFF
--- a/pkg/types/validators.go
+++ b/pkg/types/validators.go
@@ -25,6 +25,19 @@ func (v Validator) WithRunner(fn ValidateFunc) Validator {
 	return v
 }
 
+// Run applies validation to the given MetaBundle returning a result
+// indicating whether the validation was successful, failed, or encountered
+// an error.
+func (v Validator) Run(mb MetaBundle) ValidatorResult {
+	res := v.Runner(mb)
+
+	res.ValidatorCode = v.Code
+	res.ValidatorDescription = v.Description
+	res.ValidatorName = v.Name
+
+	return res
+}
+
 // ValidatorList - implements Sort interface to sort validators per Code
 type ValidatorList []Validator
 
@@ -59,20 +72,42 @@ func NewMetaBundle(addonMeta *v1alpha1.AddonMetadataSpec, bundles []registry.Bun
 	}
 }
 
+// ValidatorResultSuccess returns an otherwise empty successful validator result.
+func ValidatorResultSuccess() ValidatorResult {
+	return ValidatorResult{success: true}
+}
+
+// ValidatorResultFailure returns a failed validator result with the given failure messagees.
+func ValidatorResultFailure(msgs ...string) ValidatorResult {
+	return ValidatorResult{FailureMsgs: msgs}
+}
+
+// ValidatorResultError returns an errored validator result with the given error
+// and flag to indicate whether the error may be retried.
+func ValidatorResultError(err error, retryable bool) ValidatorResult {
+	return ValidatorResult{Error: err, RetryableError: retryable}
+}
+
 // ValidatorResult - encompasses validator result information
 type ValidatorResult struct {
 	// true if MetaBundle validation was successful
-	Success bool
+	success bool
 	// "" if validation is successful, else information about why it failed
-	FailureMsg string
-	// retports error that happened in the validation code
+	FailureMsgs []string
+	// reports error that happened in the validation code
 	Error error
 	// if an error occured in the validation code, determines if it was retryable
 	RetryableError bool
+	// ValidatorCode is the code of the Validator which produced the result instance.
+	ValidatorCode string
+	// ValidatorDescription describes the Validator which produced the result instance.
+	ValidatorDescription string
+	// ValidatorName is the name of the Validator which produced the result instance.
+	ValidatorName string
 }
 
 func (vr ValidatorResult) IsSuccess() bool {
-	return vr.Error == nil && vr.FailureMsg == ""
+	return vr.success
 }
 
 func (vr ValidatorResult) IsError() bool {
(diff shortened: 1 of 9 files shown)


pkg/types/validators.go
```diff
@@ -25,6 +25,19 @@ func (v Validator) WithRunner(fn ValidateFunc) Validator {
 	return v
 }
 
+// Run applies validation to the given MetaBundle returning a result
+// indicating whether the validation was successful, failed, or encountered
+// an error.
+func (v Validator) Run(mb MetaBundle) ValidatorResult {
+	res := v.Runner(mb)
+
+	res.ValidatorCode = v.Code
```
sugarraysam, line 34: The copying works, but if we ever want to add fields, it's a wide sync accross Validator, Run method and ValidatorResult. I am wondering if this is a temporary solution you implemented? As we had a couple of design discussion on "how to pass validator fields to middlewares"?

pkg/utils/kube_validation.go
```diff
@@ -0,0 +1,104 @@
+package utils
```
sugarraysam, line 1: I was reading how the use of catchall package names is discouraged. How about `pkg/kubeutils` ?

pkg/validate/print.go
```diff
@@ -22,29 +19,49 @@ func getTableHeaders() *simpletable.Header {
...
+func (t resultTable) WriteRow(row []*simpletable.Cell) {
+	t.Body.Cells = append(t.Body.Cells, row)
 }
 
-func newErrorTableRow(v types.Validator, err error) []*simpletable.Cell {
-	return newTableRow(v, statusError, err.Error())
+func (t resultTable) WriteResult(res types.ValidatorResult) {
+	row := resultToRow(res)
+
+	if res.IsSuccess() {
```
sugarraysam, line 39: I think this refactor has it's purpose. But the side effect is we are going over the result logic thrice now:
1. inside `WriteResult`
2. inside `resultToRow`
3. inside `ValidateCLI`

Can we go over the result logic only once?

pkg/validators/am0010_k8s_resource_names.go
```diff
@@ -20,215 +18,152 @@ func init() {
...
-		return isValidAnnotationName
-	case k8sNameLabel:
-		return isValidLabelName
-	case k8sNameNamespace:
-		return isValidNamespaceName
-	default:
-		// panic is preffered here since this indicates a programming error
-		panic(fmt.Sprintf("no validation function defined for '%v'", nameType))
+func validateBundleFields(mb types.MetaBundle) []string {
+	bp := mb.AddonMeta.BundleParameters
```
sugarraysam, line 151: The `BundleParameters` is only used for building bundles from within `managed-tenants`, and will be deprecated as soon as we migrate all bundles to `managed-tenants-bundles`.

I would discourage investing effort in validating them. The migration is ongoing https://issues.redhat.com/browse/MTSRE-162. I suppose we should add a deprecation comment to `pkg/mtsre/v1/addons.go` and `api/v1alpha1/addonmetadata_types.go`

pkg/validators/am0010_test.go
```diff
@@ -55,6 +56,13 @@ func (val TestAM0010) SucceedingCandidates() ([]types.MetaBundle, error) {
 					"foo.bar0.com/Foo-common_label.2": "true",
 					"foo.bar.com/Foo-common_label.3":  "true",
 				},
+				PagerDuty: &v1.PagerDuty{
+					SecretName:      "pagerduty-secret",
+					SecretNamespace: "redhat-foo-operator",
+				},
+				BundleParameters: &v1.BundleParameters{
```
sugarraysam, line 63: same as ^^

sugarraysam (approved):
LGTM, few nits to look into, but most importantly we should deprecate BundleParameters

pkg/types/validators.go
Ajpantuso: Yes this is temporary. My ideal state would be to inject Validators with the necessary behavior to populate results. See https://github.com/Ajpantuso/addon-metadata-operator/blob/8bbc42b81b54a2959e8b076c55b191800b5e8c44/pkg/validator/validator.go#L80-L115

pkg/utils/kube_validation.go
Ajpantuso: Yeah, `utils` is no good. Really this should be under `internal/kube...`. If it's very specifically just utilities or validations we can do `kubeutils` or anything that describes what the package is for.

pkg/validate/print.go
Ajpantuso: Uh, in a sense yes. I played around with this by having a `ResultList` which abstracts some details away https://github.com/Ajpantuso/addon-metadata-operator/blob/8bbc42b81b54a2959e8b076c55b191800b5e8c44/internal/cmd/validate.go#L102-L124. But, if you mean have a single loop where everything gets done I'm not a fan. Three loops that do one thing each is more readable and just as efficient as one loop that does three things. And if you use those `List` abstractions you can kind of make the loops disappear into the background.

pkg/validators/am0010_k8s_resource_names.go
Ajpantuso: I can put in the deprecation notices.

PR #63, opened 2022-01-17
WIP [MTSRE-404] Implementing exponential backoff retry middleware

Ajpantuso:
## Summary

Adds more configurations to existing retry middleware including exponential backoff (w/ jitter options). Also makes validators middleware pattern more similar to its inspiration `net/http`.


pkg/types/validators.go
```diff
@@ -1,28 +1,52 @@
...
+	// This is analogous to Handler.ServeHTTP
+	Run(MetaBundle) ValidatorResult
+}
+
+// ValidateFunc wraps functions which validate MetaBundles.
+type ValidateFunc func(mb MetaBundle) ValidatorResult
+
+func (f ValidateFunc) Run(mb MetaBundle) ValidatorResult {
+	return f(mb)
+}
```
sugarraysam, line 43: Do we still need this?

pkg/validators/am0001_default_channel.go
```diff
@@ -15,7 +15,7 @@ var AM0001 = types.Validator{
 	Code:        "AM0001",
 	Name:        "default_channel",
 	Description: "Ensure defaultChannel is present in list of channels",
-	Runner:      validateDefaultChannel,
+	Runner:      types.ValidateFunc(validateDefaultChannel),
```
sugarraysam, line 18: compiler will complain if `validateDefaultChannel` does not implement `ValidatorRunner` interface. Is extra cast necessary?

pkg/types/validators.go
```diff
@@ -1,28 +1,52 @@
...
-	return v
+// Validate runs a validator's Runner against a MetaBundle.
+func (v *Validator) Validate(mb MetaBundle) ValidatorResult {
+	return v.Runner.Run(mb)
+}
+
+// ValidatorRunner describes objects which run validations against
+// a MetaBundle and produce a ValidatorResult. This is analogous to
+// the http.Handler interface.
+type ValidatorRunner interface {
```
sugarraysam, line 31: We could modify `ValidatorTest` interface:

```go
type ValidatorTest interface {
        ValidatorRunner
	Name() string
	SucceedingCandidates() ([]MetaBundle, error)
	FailingCandidates() ([]MetaBundle, error)
}
```

pkg/validators/middleware.go
```diff
@@ -0,0 +1,33 @@
...
+import (
+	"github.com/mt-sre/addon-metadata-operator/pkg/types"
+)
+
+// Middleware abstracts behavior which extends types.ValidatorRunner
+// functionality by executing additional steps before/after validations
+// are run such as adding retries or logging.
+type Middleware interface {
+	// Wrap applies the implemented middleware to an instance of types.ValidatorRunner
+	Wrap(types.ValidatorRunner) types.ValidatorRunner
```
sugarraysam, line 12: An issue with this design is if we need to access `Validator` fields, for example Code or Name.

Say we want to write a `PrometheusMiddleware`, to register results, execution time and other metrics, per validator. How do we access `v.Code` or `v.Name` to add specific labels?

We would have to cast the `types.ValidatorRunner` to `types.Validator` everytime.

pkg/validators/retry.go
```diff
@@ -0,0 +1,265 @@
+package validators
```
sugarraysam, line 1: I would make it obvious in filename this is a middleware. Maybe at some point we will have a need for a separate `pkg/middlewares/`, or `pkg/validators/middleware` subpkg, but r/now maybe `retry_middleware.go` ?

pkg/validators/retry.go
```diff
@@ -0,0 +1,265 @@
...
+
+// RetryMiddleware implements middleware that conditionally
+// retries types.ValidatorRunner executions based on the result.
+type RetryMiddleware struct {
+	maxRetries int
+	delayer    types.Delayer
+}
+
+func (r *RetryMiddleware) Wrap(v types.ValidatorRunner) types.ValidatorRunner {
+	return types.ValidateFunc(func(mb types.MetaBundle) types.ValidatorResult {
```
sugarraysam, line 40: is extra `types.ValidateFunc` wrapper needed? Compiler will complain if dont respect the `types.ValidatorRunner` interface

pkg/validators/retry.go
```diff
@@ -0,0 +1,265 @@
...
+	return d2
+}
+
+func durationMax(d1, d2 time.Duration) time.Duration {
+	if d1 >= d2 {
+		return d1
+	}
+
+	return d2
+}
```
sugarraysam, line 265: :heart: +1 for exponential delay implementation. Read through quickly, I trust you got a good implementation. Off the top of your mind maybe? Rosetta code? :laughing:

pkg/validators/retry_test.go
```diff
@@ -0,0 +1,162 @@
+package validators
```
sugarraysam, line 1: would rename to `retry_middleware_test.go` following previous comment

pkg/types/validators.go
Ajpantuso: Yes, so all the questions related to typecasting `ValidateFunc` can be answered here:

Basically a `func(mb MetaBundle) ValidatorResult` cannot be given a `Run` method and will not satisfy the `ValidatorRunner` interface. So we need this additional abstraction to cast our validation functions as instances of `ValidateFunc` to satisfy the interface. It's definitely a funky Go thing.

pkg/validators/am0001_default_channel.go
Ajpantuso: See comment on `ValidateFunc`

pkg/validators/middleware.go
Ajpantuso: Yes definitely, there are two options:

1. We could follow the pattern of injecting dependencies when initializing the MW like [so](https://github.com/mt-sre/addon-metadata-operator/pull/63/files#diff-5ca607962a1e1b55993f9c37ade457ffa9aa0a8f7b55dacf5a186f597680dd98R18-R20)
2. Or we could do what you're suggesting which is a little more complex where we pass around ValidatorParams with every `ValidateFunc` like `type ValidateFunc func(params ValidatorParams, mb MetaBundle) ValidatorResult`

I think realistically we need to do the second because then we can actually pass a `context.Context`, logger, clients, etc... to the `ValidateFunc`'s themselves and then middleware will just access those params in it's closure.

pkg/validators/retry.go
Ajpantuso: See comment on `ValidateFunc`

pkg/validators/retry.go
Ajpantuso: Uh, a mix off my own knowledge and that AWS blog I mentioned in the comments :)

pkg/validators/middleware.go
Ajpantuso: Implemented the second option with latest commit

sugarraysam: I like the second options better. We have tons of flexibility to define what this `cfg` should contain too.

pkg/types/validators.go
```diff
@@ -1,28 +1,133 @@
...
+func (v *Validator) Validate(mb MetaBundle) ValidatorResult {
+	return v.Runner.Run(v.ValidatorConfig, mb)
+}
+
+// Option applies an option to an instance of Validator.
+func (v *Validator) Options(opts ...ValidatorOption) {
+	for _, opt := range opts {
+		opt(&v.ValidatorConfig)
+	}
+}
```
yashvardhan-kukreja, line 45: Just out of curiosity, any reason to make it have `Validator` as a receiver and not `ValidatorConfig` as receiver instead.

This is what I was thinking

```suggestion
func (vc *ValidatorConfig) Options(opts ...ValidatorOption) {
	for _, opt := range opts {
		opt(vc)
	}
}
```

pkg/types/validators.go
```diff
@@ -1,28 +1,133 @@
...
+	// on the implementation of a particular Delayer.
+	Delay() time.Duration
+}
+
+// Middleware abstracts behavior which extends types.ValidatorRunner
+// functionality by executing additional steps before/after validations
+// are run such as adding retries or logging.
+type Middleware interface {
+	// Wrap applies the implemented middleware to an instance of types.ValidatorRunner
+	Wrap(ValidatorRunner) ValidatorRunner
```
yashvardhan-kukreja, line 130: Loved this ❤️

pkg/types/validators.go
Ajpantuso: Nope, lol. Probably just refactored that without thinking. Your change makes perfect sense.

PR #64, opened 2022-01-18
[MTSRE-409] - Write fatal validation setup errors to stdout

Ajpantuso:
This PR writes user-friendly messages to `stdout` rather than `stderr` when using the `validate` sub-command. This change is required because output redirection does not work correctly when `mtcli` is run from CI.


sandhyadalavi (approved):
lgtm

PR #65, opened 2022-01-20
Csv rbac validations

ashishmax31:
# AM0012

## Description
This validator validates the rbac rules specified in the CSV file.

A short description of my validator and what it is meant to accomplish.
The validator currently runs the validations listed here: https://issues.redhat.com/browse/MTSRE-368.
I haven't enabled this validator yet, as this validation currently fails for many addons we handle currently.

## Checklist

- [ ] Wiki page exists: https://github.com/mt-sre/addon-metadata-operator/wiki/AMxxx
- [ ] Tested against production addons in `managed-tenants/addons/`

### DIFF
--- a/pkg/types/csv_rbac.go
+++ b/pkg/types/csv_rbac.go
@@ -3,56 +3,41 @@ package types
 import (
 	"fmt"
 	"sort"
+
+	rbac "k8s.io/api/rbac/v1"
 )
 
-type CsvPermissions struct {
-	ClusterPermissions []Permissions `json:"clusterPermissions"`
-	Permissions        []Permissions `json:"permissions"`
+type CSVPermissions struct {
+	ClusterPermissions []Permission `json:"clusterPermissions"`
+	Permissions        []Permission `json:"permissions"`
 }
 
-type Permissions struct {
-	ServiceAccount string `json:"serviceAccount"`
-	Rules          []Rule `json:"rules"`
+type Permission struct {
+	ServiceAccountName string
+	Rules              []Rule
 }
 
 type Rule struct {
-	name            string   // used in tests
-	ApiGroups       []string `json:"apiGroups"`
-	Resources       []string `json:"resources"`
-	Verbs           []string `json:"verbs"`
-	ResourceNames   []string `json:"resourceNames"`
-	NonResourceURLs []string `json:"nonResourceURLs"`
+	rbac.PolicyRule
+	name string // Used in tests
 }
-
-type operator string
-
-type permissionType string
 
 type RuleFilter struct {
-	PermissionType           permissionType
-	ApiGroupFilterObj        *FilterObj
-	ResourcesFilterObj       *FilterObj
-	VerbsFilterObj           *FilterObj
-	ResourceNamesFilterObj   *FilterObj
-	NonResourceURLsFilterObj *FilterObj
+	PermissionType permissionType
+	Filters        []Filter
 }
-
-type FilterObj struct {
+type FilterParams struct {
 	Args         []string
 	OperatorName operator
 }
 
-type filterFunc func(*Rule, RuleFilter) *Rule
+type Filter interface {
+	Filter(*rbac.PolicyRule) *rbac.PolicyRule
+}
+
+type operator string
 
 const (
-	apiGroupFilterType        = "apiGroupFilter"
-	resourcesFilterType       = "resourcesFilter"
-	verbsFilterType           = "verbsFilter"
-	resourceNamesFilterType   = "resourceNamesFilter"
-	nonResourceURLsFilterType = "nonResourceURLsFilter"
-)
-
-var (
 	InOperator           operator = "IN"
 	NotInOperator        operator = "NOT_IN"
 	EqualsOperator       operator = "EQUAL"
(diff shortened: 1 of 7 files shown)


pkg/validators/am0012_test.go
```diff
@@ -0,0 +1,47 @@
...
+func (val TestAM0012) Name() string {
+	return validators.AM0012.Name
+}
+
+func (val TestAM0012) Run(mb types.MetaBundle) types.ValidatorResult {
+	return validators.AM0012.Runner(mb)
+}
+
+func (val TestAM0012) SucceedingCandidates() ([]types.MetaBundle, error) {
+	res, err := testutils.DefaultSucceedingCandidates()
```
ashishmax31, line 25: Moved the more specific validation testing to  https://github.com/mt-sre/addon-metadata-operator/blob/a435ed8c4f8a1fc08002586ef2b01a3fea38da90/pkg/utils/csvutils/csv_rbac_utils_test.go

pkg/validators/am0012_csv_rbac.go
```diff
@@ -0,0 +1,133 @@
...
+		return Error(err)
+	}
+
+	csv, err := latestBundle.ClusterServiceVersion()
+	if err != nil {
+		return Error(err)
+	}
+	if err != nil {
+		return Error(err)
+	}
```
yashvardhan-kukreja, line 37: nit
```suggestion
	if err != nil {
		return Error(err)
	}
```

pkg/types/csv_rbac_test.go
```diff
@@ -0,0 +1,190 @@
...
+					Args:         []string{"*"},
+					OperatorName: InOperator,
+				},
+			},
+			expectedOutput: []string{"rule-2", "rule-3"},
+		},
+	}
+
+	for _, testCase := range filterCases {
+		res := inputRbacRules.FilterRules(testCase.input)
```
sugarraysam, line 183: nit: run concurrently to speed up testing  

```go
for _, tc := range filterCases {
    tc := tc // pin
    t.Run(<name>, func( t *testing.T) {
        t.Parallel()
    // ...
    })
```

pkg/utils/csvutils/csv_rbac_utils.go
```diff
@@ -0,0 +1,114 @@
...
+	for _, api := range ownedApis {
+		if api != nil {
+			result = append(result, trimWhiteSpace(api.Group))
+		}
+	}
+	return result, nil
+}
+
+func GetPermissions(csv *registry.ClusterServiceVersion) (*types.CsvPermissions, error) {
+	var objmap map[string]*json.RawMessage
```
sugarraysam, line 82: To decode the raw json, I feel it would be better to use the actual types from `operator-framework/api`: https://github.com/operator-framework/api/blob/7c97612f258921a973da23fb55dcd77892401be4/pkg/operators/v1alpha1/clusterserviceversion_types.go#L274

An example of this unmarshalling strategy is shown in AM0007: https://github.com/mt-sre/addon-metadata-operator/blob/master/pkg/validators/am0007_csv_install_modes.go#L64

Where you use a custom type.

pkg/types/csv_rbac.go
```diff
@@ -0,0 +1,263 @@
+package types
+
+import (
+	"fmt"
+	"sort"
+)
+
+type CsvPermissions struct {
```
sugarraysam, line 8: Any reason why we use custom types instead of reusing the `operator-framework/api` types? https://github.com/operator-framework/api/tree/master/pkg/operators/v1alpha1

pkg/types/csv_rbac_test.go
```diff
@@ -0,0 +1,190 @@
+package types
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestFilterRules(t *testing.T) {
```
sugarraysam, line 9: Would favor `operator-framework/api` types, but can still wrap them in the custom `CsvPermissions` wrapper type. Maybe rename to `CSVPermissions`, feels this would follow go naming philosohpy more closely.

pkg/types/csv_rbac.go
```diff
@@ -0,0 +1,263 @@
...
+	AnyOperator          operator = "ANY"
+)
+
+var (
+	AllPermissionType        permissionType = "all"
+	NameSpacedPermissionType permissionType = "namespaced"
+	ClusterPermissionType    permissionType = "clusterScoped"
+)
+
+func genFilter(attrFetcher func(*Rule) []string, filterType string) filterFunc {
```
sugarraysam, line 71: I have to admit it took me a while to understand what is happening here.

1. We define an attr fetcher to return a field of `Rule` type
2. We retrieve a FilterObj of the global `RuleFilter` struct using the custom switch statement
3. We return a new filterFunc that eventually evals the args against the filterObj

I feel the design could be simplify with clear interfaces, and contracts between objects.

For example, the `FilterObj` global struct really feels like different objects that should respect the `RuleFilter` interface. E.g:

```go
type RuleFilter interface {
    Filter(rule *Rule, args []string, op operator) *Rule
}

type ApiGroupFilter struct {}

func (f *ApiGroupFilter) Filter(rule *Rule, args []string, op operator) *Rule {
    return eval(rule.ApiGroups, args, op)
}
```

But IDK im just writing this quickly, maybe our interface expert @Ajpantuso would have some ideas.

Also I would dissociate the `PermissionType` from the `FilterObj`. From a filter's perspective, it should not care what rules it receives, as long as they are rules. Some external obj should determine if we should pass `clusterPermissions` or `permissions` or both.

sugarraysam (commented):
This is good work. Although I feel the design could be revisited by adding interfaces and defining clear boundaries between objects.

pkg/types/csv_rbac.go
ashishmax31: @sugarraysam Sure, let me try to refactor it, thanks for the feedback.
> Also I would dissociate the PermissionType from the FilterObj
 Yeah now that you've written it down, it makes total sense. :D

pkg/types/csv_rbac.go
ashishmax31: I thought of creating a new type so that we can add new attributes in the future if needed. But like you said, wrapping the `operator-framework/api` type inside this custom struct also works. Will refactor this PR.

ashishmax31:
> This is good work. Although I feel the design could be revisited by adding interfaces and defining clear boundaries between objects.

Thanks for the feedback Sam, I'll refactor the code to use interfaces.

ashishmax31:
@sugarraysam Please have a look now

Ajpantuso (commented):
Solid test coverage and abstraction to keep the business logic clean. Instead of suggestions here I made an example [commit](https://github.com/Ajpantuso/addon-metadata-operator/commit/092da65073fdc74147e4bc553d752a31ccebf6bc) to demonstrate some ways of simplifying the filters a bit and reducing boilerplate. Please take a look and see if it would be an improvement.

ashishmax31:
yep, checking it out. thanks @Ajpantuso

ashishmax31:
@Ajpantuso @sugarraysam  Made the requested changes please have a look now.

Ajpantuso (approved):
LGTM, thanks for working on this!

PR #66, opened 2022-01-20
[MTSRE-405] Add validation_err_msg to AddOnParameters schema.

sugarraysam:
Also modify AM0009 to use the new user provided error message.


Ajpantuso (approved):
LGTM

PR #67, opened 2022-01-24
AM0007: Validate installMode type in metadata

mayankshah1607:
Update AM0007 to also validate the installMode type in the metadata. This check ensures that the metadata only supports AllNamespaces and OwnNamespace install mode now that SingleNamespace has been completely dropped from managed-tenants.

### DIFF
--- a/pkg/validators/am0007_csv_install_modes.go
+++ b/pkg/validators/am0007_csv_install_modes.go
@@ -21,8 +21,24 @@ var AM0007 = types.Validator{
 	Runner:      validateCSVInstallModes,
 }
 
+var validInstallModes = []string{"AllNamespaces", "OwnNamespace"}
+
+func indexOf(element string, data []string) int {
+	for k, v := range data {
+		if element == v {
+			return k
+		}
+	}
+	return -1 //not found.
+}
+
 func validateCSVInstallModes(mb types.MetaBundle) types.ValidatorResult {
 	installMode := mb.AddonMeta.InstallMode
+
+	// allow only AllNamespaces and OwnNamespace install mode.
+	if indexOf(installMode, validInstallModes) == -1 {
+		return Fail(fmt.Sprintf("unsupported install mode %v", installMode))
+	}
 
 	for _, bundle := range mb.Bundles {
 		bundleName, err := utils.GetBundleNameVersion(bundle)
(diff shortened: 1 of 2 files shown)


pkg/validators/am0007_csv_install_modes.go
```diff
@@ -21,20 +21,36 @@ var AM0007 = types.Validator{
...
+	}
+	return -1 //not found.
+}
+
 func validateCSVInstallModes(mb types.MetaBundle) types.ValidatorResult {
 	installMode := mb.AddonMeta.InstallMode
 
+	// allow only AllNamespaces and OwnNamespace install mode.
+	if indexOf(installMode, validInstallModes) == -1 {
+		return Error(fmt.Errorf("unsupported install mode %v", installMode))
```
yashvardhan-kukreja, line 40:
```suggestion
		return Fail(fmt.Sprintf("unsupported install mode %v", installMode))
```

yashvardhan-kukreja:
@mayankshah1607 all looks good to me except one question.

In this migration, are we fine with just ensuring that `spec.InstallMode` is one of AllNamespaces/OwnNamespace or do we also want to enforce that even under `spec.InstallModes`, all the items correspond to only AllNamespaces/OwnNamespace?

mayankshah1607:
@yashvardhan-kukreja For now I think it's enough to just validated that the InstallMode in the metadata is one of AllNamespace/OwnNamespace

yashvardhan-kukreja (approved):

yashvardhan-kukreja:
I'd suggest that for the very least, we should enforce that under `spec.InstallModes`, there shouldn't be `supported: true` against any mode apart from `AllNamespaces/OwnNamespaces`.
For that, we can edit the following method to this: 

```go
func isInstallModeSupported(installModes []operatorsv1alpha1.InstallMode, target string) (bool, string) {
	var allSupported []string
	targetSupported := false
	for _, im := range installModes {
		imType := string(im.Type)
		if im.Supported {
                         if indexOf(im.Type, validInstallModes) == -1 {
		                return false, fmt.Sprintf("One of the installModes %v is not supported. CSV only supports these installModes %v.", imType, validInstallModes)
	                }
                        allSupported = append(allSupported, imType)
		}
		if imType == target {
			targetSupported = im.Supported
		}
	}
	return targetSupported, fmt.Sprintf("Target installMode %v is not supported. CSV only supports these installModes %v.", target, allSupported)
}
```

I am happy with this PR too so it's your call to create a new PR for this suggestion or update this PR itself :)

mayankshah1607:
> there shouldn't be supported: true against anytype apart from AllNamespaces/OwnNamespaces

We could have this check, but it doesn't really matter. The metadata will not allow `SingleNamespace`, so if the CSV contains `supported: true` for `SingleNamespace`, the mismatch will be caught by this validator anyway.

PR #69, opened 2022-01-25
[MTSRE-376] Replace the old extractor. 

sugarraysam:
Also an important change is that we are going to be passing bundle using pointers as they are expensive to copy, and we only need read-access. Bundles are meant to be immutable.


pkg/extractor/index_test.go
```diff
@@ -74,43 +75,39 @@ func TestIndexExtractorListAllBundles(t *testing.T) {
...
-// cache or indexImage list bundles in different order, using a map to avoid flakiness
-func allBundleImagesListed(bundleImages []string, expectedBundleImages map[string]bool) bool {
-	if len(bundleImages) != len(expectedBundleImages) {
-		return false
-	}
-	for _, image := range bundleImages {
-		if _, ok := expectedBundleImages[image]; !ok {
-			return false
-		}
+func requireContainsAll(t *testing.T, images []string, expectedImages []string) {
```
Ajpantuso, line 108: I think`testify` already has this https://pkg.go.dev/github.com/stretchr/testify/require#Assertions.ElementsMatch
Except `ElementsMatch` would fail for `images=[]string{"1","1","1"} expectedImages=[]string{"1","2","3"}` whereas this would not.

pkg/types/validators.go
```diff
@@ -49,10 +49,10 @@ type ValidatorTest interface {
 
 type MetaBundle struct {
 	AddonMeta *v1alpha1.AddonMetadataSpec
-	Bundles   []registry.Bundle
+	Bundles   []*registry.Bundle
```
Ajpantuso, line 52: You should only be passing `registry.Bundle` pointers to functions for performance benefits. Using slices of pointers has the opposite effect https://philpearl.github.io/post/bad_go_slice_of_pointers. The TL;DR is a `[]struct{Field}` allows for allocation of a contiguous block of `struct{}` whereas a `[]*struct{Field}` only allows for allocation of a contiguous block of `*struct{Field}` meaning the allocation for the memory addressed by those pointers is defferred.

Ajpantuso (changes requested):
Couple of comments, but otherwise looks good.

sugarraysam:
Can't merge before we fix non-OPM compliance: https://issues.redhat.com/browse/MTSRE-382 ... :(

pkg/extractor/index_test.go
sugarraysam: Was looking for that. Indeed my testing helper had a bug :+1:

pkg/types/validators.go
sugarraysam: I see, as slices are always passed by reference, it's not an expensive copy to pass `[]registry.Bundle`, and then individual function use `*registry.Bundle` to avoid the expensive copy of the bundle.

sugarraysam: Decided to run some benchmark (`$ make benchmark`). I don't see a huge difference actually:

sugarraysam: Let me take a second stab at those benchmarks

sugarraysam: New results PTAL @Ajpantuso : 

```bash
cpu: Intel(R) Core(TM) i7-10850H CPU @ 2.70GHz
BenchmarkBundlesByReference
BenchmarkBundlesByReference-12               224           5584239 ns/op         2429805 B/op      18882 allocs/op
BenchmarkBundlesByReference-12               208           5764610 ns/op         2430565 B/op      18882 allocs/op
BenchmarkBundlesByReference-12               169           6351222 ns/op         2431537 B/op      18882 allocs/op
BenchmarkBundlesByReference-12               192           6518670 ns/op         2431161 B/op      18882 allocs/op
BenchmarkBundlesByReference-12               169           6267681 ns/op         2430601 B/op      18882 allocs/op
BenchmarkBundlesByReference-12               198           5827175 ns/op         2428280 B/op      18881 allocs/op
BenchmarkBundlesByReference-12               204           5672996 ns/op         2433242 B/op      18883 allocs/op
BenchmarkBundlesByReference-12               175           7277202 ns/op         2430600 B/op      18882 allocs/op
BenchmarkBundlesByReference-12               207           6110458 ns/op         2433272 B/op      18883 allocs/op
BenchmarkBundlesByReference-12               100          10184805 ns/op         2428210 B/op      18882 allocs/op
BenchmarkBundlesByValue
BenchmarkBundlesByValue-12                   100          10311958 ns/op         2429499 B/op      18882 allocs/op
BenchmarkBundlesByValue-12                   190           7846127 ns/op         2431620 B/op      18882 allocs/op
BenchmarkBundlesByValue-12                   202           6890022 ns/op         2430242 B/op      18882 allocs/op
BenchmarkBundlesByValue-12                   172           6143083 ns/op         2430244 B/op      18882 allocs/op
BenchmarkBundlesByValue-12                   166           8560888 ns/op         2428309 B/op      18880 allocs/op
BenchmarkBundlesByValue-12                   153           6820581 ns/op         2430576 B/op      18882 allocs/op
BenchmarkBundlesByValue-12                   170           7265319 ns/op         2430702 B/op      18881 allocs/op
BenchmarkBundlesByValue-12                   159           7549030 ns/op         2431304 B/op      18882 allocs/op
BenchmarkBundlesByValue-12                   158           8021400 ns/op         2430546 B/op      18881 allocs/op
BenchmarkBundlesByValue-12                   180           6706331 ns/op         2434760 B/op      18882 allocs/op
```

Ajpantuso: Yeah, so now the allocations just represent what each iteration of the `NewBundle` function requires rather than populating the slice. The deoptimization would become clear if for the `[]registry.Bundle` test `NewBundle` was a function which took a `*registry.Bundle` as an arg and modified the value it points to. So the benchmark shows that for iterative bundle loading we would see about equal performance. What it doesn't show is that if we ever wanted to copy `registry.Bundle` values from one slice to another, i.e. create a new underlying array, using pointer slices would slow us down.

sugarraysam: Unfortunately, I do not understand what you mean. My objective was to replicate what our workflow looks like.

1. Extract bundles and buildup a slice (extractor)
2. Pass the slice around (create new metabundle obj)
3. Iterate through the bundles (validators)

I've added another line where we create a dummy metabundle to represent our workflow more tightly.

sugarraysam: https://github.com/mt-sre/addon-metadata-operator/pull/69/files#diff-c3936c48f3bb68586d6f4bd4bed5e325ceeb6adf92dea4b1d04fe5035e8cc1f2

Ajpantuso: Current workflow is fine, but if the goal was performance then choosing to use `[]*registry.Bundle` introduces operations that we'd have to avoid. Tried to remove any excess from these examples:

```go
func BenchmarkBundleValueSliceCopy(b *testing.B) {
	b.ReportAllocs()

	for i := 0; i < b.N; i++ {
		var bundles []registry.Bundle

		copy(bundles, bundleValueSlice())
	}
}

func BenchmarkBundlePtrSliceCopy(b *testing.B) {
	b.ReportAllocs()

	for i := 0; i < b.N; i++ {
		var bundles []*registry.Bundle

		copy(bundles, bundlePtrSlice())
	}
}

func bundleValueSlice() []registry.Bundle {
	bundles := make([]registry.Bundle, 0, 1)

	bundles = append(bundles, registry.Bundle{})

	return bundles
}

func bundlePtrSlice() []*registry.Bundle {
	bundles := make([]*registry.Bundle, 0, 1)

	bundles = append(bundles, &registry.Bundle{})

	return bundles
}
```

Value gives:
```bash
goos: linux
goarch: amd64
pkg: github.com/mt-sre/addon-metadata-operator/pkg/extractor
cpu: Intel(R) Core(TM) i7-10875H CPU @ 2.30GHz
BenchmarkBundleValueSliceCopy-16    	100000000	        11.13 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	github.com/mt-sre/addon-metadata-operator/pkg/extractor	1.149s
```

Ptr gives:
```bash
goos: linux
goarch: amd64
pkg: github.com/mt-sre/addon-metadata-operator/pkg/extractor
cpu: Intel(R) Core(TM) i7-10875H CPU @ 2.30GHz
BenchmarkBundlePtrSliceCopy-16    	14999680	        78.74 ns/op	     240 B/op	       1 allocs/op
PASS
ok  	github.com/mt-sre/addon-metadata-operator/pkg/extractor	1.284s
```

The one extra allocation means anytime a slice of bundles needs to be copied or appended to, even with preallocation, the task is 7x slower.

Ajpantuso: This is all it's about:
```go
func BenchmarkBundleValueOutsideAllocation(b *testing.B) {
	b.ReportAllocs()

	for i := 0; i < b.N; i++ {
		bundles := make([]registry.Bundle, 0, 1)

		bundle := new(registry.Bundle)

		bundles = append(bundles, *bundle) // bundle stays on the stack
	}
}

func BenchmarkBundlePtrOutsideAllocation(b *testing.B) {
	b.ReportAllocs()

	for i := 0; i < b.N; i++ {
		bundles := make([]*registry.Bundle, 0, 1)

		bundle := new(registry.Bundle)

		bundles = append(bundles, bundle) // bundle escapes to the heap incurring one allocation
	}
}
```

So if we never initialize our own `registry.Bundle` objects with functions which can be inlined then it's probably not an issue.

Ajpantuso:
Closing in favor of #113

PR #81, opened 2022-03-03
Add the mtcli bundle validate command to validate local bundle dirs.

sugarraysam:
Creating a new `mtcli validate-bundle <bundle_path>` command to mimic `opm alpha bundle validate --tag <image>` but using local directory path.


sandhyadalavi:
can we add the test? it will be very similar to this? https://github.com/mt-sre/addon-metadata-operator/blob/master/integration/mtcli/list_bundles_test.go

sugarraysam:
Thanks for your feedback @sandhyadalavi . I think Im actually going to trash this PR, and tackle the problem at the source, meaning write a new containerregistry implementation that first looks for locally built images, will also solve the `list-bundles` issue.

sugarraysam:
Changes !
- added a bunch of testdata: `internal/testdata/bundles/*`
- refactored cli to have subcommands `mtcli bundle [subcommand]` + `mtcli list [subcommand]`
- added e2e test

sandhyadalavi (approved):
lgtm

PR #84, opened 2022-03-08
[MTSRE-472] Extending operatorName valid length

Ajpantuso:
### Summary

Increases the validation length for `operatorName` beyond 32 characters.


sugarraysam (approved):
LGTM

PR #87, opened 2022-03-11
Scaffolding AMO operator and testdata.

sugarraysam:
## Description

Adding operator scaffolding for integration testing, and refactored our testdata to add concept of metadata v1 vs. v2 (i.e.: CRD-based or not).

## Changes

- Change CRD group to `addonsflow.redhat.openshift.io`: https://coreos.slack.com/archives/C01L83EMDUZ/p1641821371028100
- Reorganized `internal/testdata/`, view `internal/testdata/README.md`
- Added reference-addon CRDs => `internal/testdata/metadata_v2/reference-addon/` ++ successfully `kubectl apply -R -f internal/testdata/metadata_v2/` onto kind cluster
- Setup kind cluster for integration tests: `$ make kind-integration`
- Modify all instances of `path.Join` -> `filepath.Join` for OS compatibility


sandhyadalavi:
are we going to make any changes to use mtcli? or we will just as python lib?

sandhyadalavi:
are we going to make any changes to use mtcli? or we will just as python lib?

sugarraysam:
I am not sure I understand. This is scaffolding to help building the addon-metadata-operator. It is separate from mtcli. The codebase produces 2 artifacts:
- mtcli binary
- addon-metadata-operator binary/container

api/v1alpha1/addonimageset_utils.go
```diff
@@ -0,0 +1,32 @@
...
+	"encoding/json"
+	"fmt"
+	"strings"
+
+	"golang.org/x/mod/semver"
+	"k8s.io/apimachinery/pkg/util/yaml"
+)
+
+// FromYAML - instantiates an AddonImageSetSpec struct from yaml data
+func (a *AddonImageSetSpec) FromYAML(data []byte) error {
```
sandhyadalavi, line 13: I noticed that we've ta est for GetSemver but not for FromYAML or ToJSON, we should add it.

api/v1alpha1/addonmetadata_utils.go
```diff
@@ -0,0 +1,53 @@
...
+
+import (
+	"encoding/json"
+
+	ocmv1 "github.com/mt-sre/addon-metadata-operator/pkg/ocm/v1"
+	"k8s.io/apimachinery/pkg/util/yaml"
+)
+
+// FromYAML - instantiates an AddonMetadataSpec struct from yaml data
+func (a *AddonMetadataSpec) FromYAML(data []byte) error {
```
sandhyadalavi, line 11: same for these funcs too.

internal/controllers/addonmetadata/controller.go
```diff
@@ -22,41 +22,49 @@ import (
...
+	meta := &addonsv1alpha1.AddonMetadata{}
+	if err := r.Get(ctx, req.NamespacedName, meta); err != nil {
+		return ctrl.Result{}, client.IgnoreNotFound(err)
+	}
+
+	data, err := meta.ToJSON()
+	if err != nil {
+		log.Error(err, "could not marshal meta CR to JSON")
+	}
+	log.Info(string(data))
```
sandhyadalavi, line 63: we dont want to log for every reconcile loop

api/v1alpha1/addonimageset_utils.go
sugarraysam: well both FromYAML and ToJSON only really use the `json` and `yaml` modules. A test would not cover anything that is owned by our codebase. These are just stylistic helpers so we can do:

```go
data, err := addonmetadata.ToJSON()

// instead of
data, err := json.Marshal(meta)
```

internal/controllers/addonmetadata/controller.go
sugarraysam: this is just dummy code for now, will be wiped out by another PR. Basically this PR is committing all the required scaffolding, after which we can start focusing on reconciliation logic.

sandhyadalavi (approved):
lgtm

PR #88, opened 2022-03-15
Fixing `AM0012` false failures for newly on-boarded addons

Ajpantuso:
### Summary

Prevents false failures when a new addon is on-boarding and does not have any bundles yet.


sandhyadalavi (approved):
lgtm

PR #89, opened 2022-03-16
[MTSRE-287] Updating from SkuRules endpoint to QuotaRules

Ajpantuso:
### Summary

Using the `QuotaRules` endpoint to permit validation of addons with zero `ocmQuotaCost` defined.


sandhyadalavi (approved):
lgtm

PR #91, opened 2022-03-22
Removing root permissions from build_tag.sh

Ajpantuso:
### Summary

- Removes root permissions from `build_tag.sh`
- Removes dependency on host volume when publishing release artifacts
- Moves release logic to `Makefile`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@
 .SHELLFLAGS=-euo pipefail -c
 SHELL := /bin/bash
 
-.PHONY: all test fmt vet clean build tidy docker-build docker-push test-e2e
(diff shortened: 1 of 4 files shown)


sandhyadalavi (approved):
lgtm

Makefile
```diff
@@ -63,6 +63,19 @@ docker-push: ## Push docker image with the operator.
 	@docker --config=$(DOCKER_CONF) push $(REPO):$(TAG)
 	@docker --config=$(DOCKER_CONF) push $(REPO):latest
 
+##@ Release
```
sugarraysam, line 66: Wondering why we need all these build targets? Seems this is simply the `build_tag.sh` implementation. It does not really matter but I think bash is easier to read than 3 makefile targets.

In any case thanks for the fix LGTM

sugarraysam (approved):

PR #92, opened 2022-03-30
Validator Refactor

Ajpantuso:
### Summary

Prepares validators for long-lived runtimes by passing dependencies and context from the initializing runner. Also helps contain import side-effects by pulling out the side-effectual bit to an isolated package (making testing much easier).


pkg/validator/runner.go
```diff
@@ -0,0 +1,222 @@
...
+		}(val)
+	}
+
+	go func() {
+		wg.Wait()
+		close(resultCh)
+	}()
+
+	return resultCh
+}
```
sugarraysam, line 107: Fire fire :+1: 

We did not use any multithreading before if I remember correctly.

pkg/validator/runner.go
```diff
@@ -0,0 +1,222 @@
...
+	}
+}
+
+func (c *RunnerConfig) Default() error {
+	if len(c.Initializers) == 0 {
+		c.Initializers = initializers
+	}
+
+	if c.Logger == nil {
+		c.Logger = logr.Discard()
```
sugarraysam, line 147: Can we default to the info logger instead?

pkg/validator/runner.go
```diff
@@ -0,0 +1,222 @@
...
+type Filter func(Validator) bool
+
+func MatchesCodes(codes ...Code) Filter {
+	return func(v Validator) bool {
+		for _, c := range codes {
+			if v.Code() != c {
+				continue
+			}
+
+			return true
```
sugarraysam, line 211: nit:

```go
if v.Code() == c {
  return true
}
```

sugarraysam (approved):
LGTM, tons of improvements from what we had before. I like the reorganization. Good refactors on the filter object, new testing utils, context and dependency injection. Really good stuff thank you for this !!

pkg/validator/runner.go
Ajpantuso: The only issue with that is that there is no default `logr` implementation so we'd have to pick one. The one benefit to defaulting to `Discard` is that it produces no side-effects which is nice for testing.

pkg/validator/register/register.go
```diff
@@ -0,0 +1,15 @@
...
+	_ "github.com/mt-sre/addon-metadata-operator/pkg/validator/am0004"
+	_ "github.com/mt-sre/addon-metadata-operator/pkg/validator/am0005"
+	_ "github.com/mt-sre/addon-metadata-operator/pkg/validator/am0006"
+	_ "github.com/mt-sre/addon-metadata-operator/pkg/validator/am0007"
+	_ "github.com/mt-sre/addon-metadata-operator/pkg/validator/am0008"
+	_ "github.com/mt-sre/addon-metadata-operator/pkg/validator/am0009"
+	_ "github.com/mt-sre/addon-metadata-operator/pkg/validator/am0010"
+	_ "github.com/mt-sre/addon-metadata-operator/pkg/validator/am0011"
+	_ "github.com/mt-sre/addon-metadata-operator/pkg/validator/am0012"
+)
```
sugarraysam, line 15: Should we document this new init process. If I understand correctly, the idea is to dummy import this register package, which dummy imports all validators, so that all init funcs are ran? Probably need to review the whole `validators_test.md` doc.

pkg/validators/middlewares.go
```diff
@@ -1,51 +0,0 @@
-package validators
```
sugarraysam, line 1: Realized we are wiping the concept of middlewares completly. Is this intended?

pkg/validator/runner.go
```diff
@@ -0,0 +1,220 @@
...
+// Register queues an "Initializer" which the "Runner"
+// will invoke upon startup.
+func Register(init Initializer) {
+	initializers = append(initializers, init)
+}
+
+// Initializer is a function which will initialize a Validator
+// with dependencies. An error is returned if the Validator
+// cannot be initalized properly.
+type Initializer func(Dependencies) (Validator, error)
```
sugarraysam, line 25: Is this the replacement for middlewares? How can we migrate the retry-logic middleware? Or inject a prometheus reporter middleware for each validator?

pkg/validators/middlewares.go
Ajpantuso: For the moment yes (See the next comment).

pkg/validator/runner.go
Ajpantuso: Not really. Middleware can still be added with a signature like `type Middleware func(Validator) Validator` and any Middleware would be injected as a `RunnerOption`. The `Runner` would refer to a `[]Middleware` during its `Run` function to wrap each Validator. That way you can have pre/post task execution like timers or metric collection. Anything more granular like debug logging would be injected from the `Initializer` func.

pkg/validator/runner.go
sugarraysam: OK, this is not a big deal r/now, it is pluggable in any case. We will want to distinguish CLI from k8s environment:
- CLI inherits from a global `--log-level` flag or something but can default to Discard
- k8s (as kube jobs) default to at least Info, with flag option on operator to reach `Debug`

sugarraysam:
LGTM after middleware review and feature parity. good to go

sugarraysam:
LGTM after middleware review and feature parity. good to go

PR #100, opened 2022-04-26
ci: integrate common go-ci tools

Ajpantuso:
### Summary

Update `magefile.go` to use common tooling where possible.


magefile.go
```diff
@@ -61,79 +70,177 @@ var Aliases = map[string]interface{}{
...
+		command.WithCurrentEnv(true),
+		command.WithEnv{
+			"CGO_CFLAGS": "-DSQLITE_ENABLE_JSON1",
+		},
+		command.WithArgs{
+			"test", "-count=1", "-race", "-timeout", "15m",
+			"./api...",
+			"./cmd...",
+			"./internal...",
+			"./pkg...",
```
sugarraysam, line 177: Is there a way to reverse that logic. If we can simply exclude `integration` we don't have to maintain that list for new packages/dirs?

magefile.go
```diff
@@ -147,260 +254,341 @@ func (Test) Clean() error {
...
+
+	if err := generate.Run(); err != nil {
+		return fmt.Errorf("starting to generate boilerplate: %w", err)
+	}
+
+	if generate.Success() {
+		return nil
+	}
+
+	return fmt.Errorf("generating boilerplate: %w", generate.Error())
```
sugarraysam, line 397: nit: this seems like a repeatable pattern that might be abstracted. Just brainstorming but we might avoid the extra Success/Error check by modifying the Run method of the command interface: 

```go
// Run executes a "Command" instance and returns an error if
// either the command was not able to start.
func (c *Command) Run() error {
	if err := c.cmd.Run(); err != nil {
		if _, ok := err.(*exec.ExitError); !ok { //nolint:errorlint
			return fmt.Errorf("running command %q: %w", strings.Join(c.cmd.Args, " "), err)
		}
	}

	if !c.cmd.ProcessState.Success() {
            return fmt.Errof("command return error %q: %w", strings.Join(c.cmd.Args, " "), c.cmd.ProcessState.String())
        }

        return nil
        }
}
```

sugarraysam: (not going to paste this comment all over but pretty much all commands use that pattern)

Ajpantuso: The issue is there still needs to be a way to distinguish between a failure to launch (e.g. a missing binary) or a process failure. The examples in this `magefile` are a little expanded because I'm wrapping the error.

So this:
```
if generate.Success() {
	return nil
}

return fmt.Errorf("generating boilerplate: %w", generate.Error())
```

Could just be:
```
return generate.Error()
```

This was actually one of the issues with the `magefile/sh` abstractions. It was difficult to react to failures since every error looks the same.

sugarraysam: That makes sense. Although you could convey the difference in the error message returned by `Run`

sugarraysam: I just feel as a developer/user of this CI library, I don't really care about the error type, as long as it's providing enough information for me to debug. I don't see the value in the extra error splitting complexity. But that's just opinion.

sugarraysam: I just feel as a developer/user of this CI library, I don't really care about the error type, as long as it's providing enough information for me to debug. I don't see the value in the extra error splitting complexity. But that's just opinion.

sugarraysam (approved):
/lgtm

PR #103, opened 2022-04-29
ci: goreleaser config enhancements

Ajpantuso:
### Summary

Enhancements to the `.goreleaser.yml` file which does the following:

- Eliminates duplicated archives caused by an superfluous list item
- Reduces duplication with YAML anchors
- Removes dead code for `*386` builds
- Renames builds to reflect compiler rather than target
- Makes the changelog compatible with [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/)

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -6,59 +6,62 @@
 # Validate config with: $ goreleaser check
 project_name: mtcli
 builds:
-  - id: linux-amd64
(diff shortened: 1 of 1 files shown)


sugarraysam (approved):
/lgtm  for commit conventions I am all for it. There is some dev awareness work to be done.

PR #104, opened 2022-04-29
docs: dev docs for adding new validators

Ajpantuso:
### Summary

Updating documentation for adding new validators.


docs/adding_validators.md
```diff
@@ -0,0 +1,86 @@
+# Adding Validators
+
+## Overview
+
+New validation logic for addon metadata may require a new validator to be
```
apahim, line 5: "may require" is a bit confusing. Suggestion:

Before creating a new validator, check if the logic should fit an existing validator.

In case you create a new validator, it has to be added to the...

apahim (approved):
Looks awesome. Minor nit, safe to ignore :)

docs/adding_validators.md
Ajpantuso: Reworded it a bit if you want to check again.

PR #111, opened 2022-05-30
Refactoring AM0003 code according to the current directory structure.

Ankit152:
# AM0003

## Description

Refactoring Sam's PR for `AM0003` validator according to the current directory structure of `AMO`.

PR #140, opened 2022-07-15
[MTSRE-390] AMO: validate annotations on AM0001 and AM0003

Ankit152:

## Description

Added validation for annotations for validators AM0001 (AM0003 is already covered in PR #111).
## Checklist

- [x] Wiki page exists: https://github.com/mt-sre/addon-metadata-operator/wiki/AM0001
- [x] Tested against production addons in `managed-tenants/addons/`

### DIFF
--- a/pkg/validator/am0001/default_channel.go
+++ b/pkg/validator/am0001/default_channel.go
@@ -3,10 +3,12 @@ package am0001
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"github.com/mt-sre/addon-metadata-operator/api/v1alpha1"
 	"github.com/mt-sre/addon-metadata-operator/pkg/types"
 	"github.com/mt-sre/addon-metadata-operator/pkg/validator"
+	"github.com/operator-framework/operator-registry/pkg/registry"
 )
 
 func init() {
@@ -50,6 +52,10 @@ func (d *DefaultChannel) Run(ctx context.Context, mb types.MetaBundle) validator
 		return res
 	}
 
+	if res := d.matchesBundleChannelAnnotations(defaultChannel, mb.Bundles); !res.IsSuccess() {
+		return res
+	}
+
 	return d.Success()
 }
 
@@ -83,4 +89,47 @@ func (d *DefaultChannel) isListedInChannels(channels *[]v1alpha1.Channel, defaul
 	}
 	msg := fmt.Sprintf("The defaultChannel '%v' is not part of the listed channelNames: %v.", defaultChannel, channelNames)
 	return d.Fail(msg)
+}
+
+func (d *DefaultChannel) matchesBundleChannelAnnotations(defaultChannel string, bundles []*registry.Bundle) validator.Result {
+	var message []string
+	bundle, err := validator.GetLatestBundle(bundles)
+	if err != nil {
+		return d.Fail("Error while checking bundles")
+	}
+
+	if bundle.Annotations.DefaultChannelName == "" && defaultChannel != "alpha" {
+		msg := fmt.Sprintf("operators.operatorframework.io.bundle.channel.default.v1 is not defined so defaultChannel should be 'alpha' instead of '%v'", defaultChannel)
+		message = append(message, msg)
+	}
+
+	if bundle.Annotations.DefaultChannelName != defaultChannel && bundle.Annotations.DefaultChannelName != "" {
+		msg := fmt.Sprintf("The defaultChannel '%v' does not match annotation operators.operatorframework.io.bundle.channel.default.v1 '%v'.",
+			defaultChannel, bundle.Annotations.DefaultChannelName,
+		)
+		message = append(message, msg)
+	}
+
+	channels := strings.Split(bundle.Annotations.Channels, ",")
+
+	if !isPresentInBundleChannels(defaultChannel, channels) {
+		msg := fmt.Sprintf("The defaultChannel '%v' is not present in annotation operators.operatorframework.io.bundle.channels.v1 '%v'.",
+			defaultChannel, channels,
+		)
+		message = append(message, msg)
+	}
+
+	if len(message) > 0 {
+		return d.Fail(message...)
+	}
+	return d.Success()
+}
+
+func isPresentInBundleChannels(defaultChannel string, channels []string) bool {
+	for _, channel := range channels {
+		if channel == defaultChannel {
+			return true
+		}
+	}
+	return false
 }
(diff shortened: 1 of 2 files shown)


pkg/validator/am0001/default_channel.go
```diff
@@ -84,3 +90,38 @@ func (d *DefaultChannel) isListedInChannels(channels *[]v1alpha1.Channel, defaul
...
+		if bundle.Annotations.DefaultChannelName != defaultChannel && bundle.Annotations.DefaultChannelName != "" {
+			msg := fmt.Sprintf("The defaultChannel is '%v' is not matching operators.operatorframework.io.bundle.channel.default.v1 '%v'.",
+				defaultChannel, bundle.Annotations.DefaultChannelName,
+			)
+			message = append(message, msg)
+		}
+
+		channels := strings.Split(bundle.Annotations.Channels, ",")
+
+		isPresentInBundleChannels := func(defaultChannel string, channels []string) bool {
```
garrettthomaskth, line 106: Here is a good article about inner functions https://realpython.com/inner-functions-what-are-they-good-for/#using-inner-vs-private-helper-functions. It is about python but I think the same applies to golang. It also says when it is just a helper function, it is probably better to define it outside the other function.

garrettthomaskth: Also, the function is only called once, so I don't necessarily think the code has to be pulled out into a different function :)

pkg/validator/am0001/default_channel.go
```diff
@@ -84,3 +90,38 @@ func (d *DefaultChannel) isListedInChannels(channels *[]v1alpha1.Channel, defaul
 	msg := fmt.Sprintf("The defaultChannel '%v' is not part of the listed channelNames: %v.", defaultChannel, channelNames)
 	return d.Fail(msg)
 }
+
+func (d *DefaultChannel) isListedInBundleChannel(defaultChannel string, bundles []*registry.Bundle) validator.Result {
+	var message []string
+	for _, bundle := range bundles {
+		if bundle.Annotations.DefaultChannelName != defaultChannel && bundle.Annotations.DefaultChannelName != "" {
+			msg := fmt.Sprintf("The defaultChannel is '%v' is not matching operators.operatorframework.io.bundle.channel.default.v1 '%v'.",
```
garrettthomaskth, line 98:
```suggestion
			msg := fmt.Sprintf("The defaultChannel '%v' does not match the annotation operators.operatorframework.io.bundle.channel.default.v1 '%v'. ",
```

pkg/validator/am0001/default_channel.go
```diff
@@ -84,3 +90,38 @@ func (d *DefaultChannel) isListedInChannels(channels *[]v1alpha1.Channel, defaul
...
+				defaultChannel, bundle.Annotations.DefaultChannelName,
+			)
+			message = append(message, msg)
+		}
+
+		channels := strings.Split(bundle.Annotations.Channels, ",")
```
garrettthomaskth, line 104: Can we be 100% sure that there are no spaces in the annotation? Would it be safer to do `strings.Split(strings.ReplaceAll(bundle.Annotations.Channels," ","") , ",")`

pkg/validator/am0001/default_channel.go
```diff
@@ -84,3 +90,38 @@ func (d *DefaultChannel) isListedInChannels(channels *[]v1alpha1.Channel, defaul
...
+			for _, channel := range channels {
+				if defaultChannel == channel {
+					return true
+				}
+			}
+			return false
+		}
+
+		if !isPresentInBundleChannels(defaultChannel, channels) {
+			msg := fmt.Sprintf("The defaultChannel is '%v' is not present in operators.operatorframework.io.bundle.channels.v1 '%v'",
```
garrettthomaskth, line 116:
```suggestion
			msg := fmt.Sprintf("The defaultChannel '%v' is not present in operators.operatorframework.io.bundle.channels.v1 '%v'. ",
```

pkg/validator/am0001/default_channel_test.go
```diff
@@ -58,5 +59,24 @@ func TestDefaultChannelInvalid(t *testing.T) {
...
+			},
+			Bundles: []*registry.Bundle{
+				{
+					Annotations: &registry.Annotations{
+						DefaultChannelName: "alpha",
+						Channels:           "beta,stable,rc",
+					},
+				},
+			},
+		},
```
garrettthomaskth, line 80: You could also add one more test testing the validation if the `DefaultChannel` field doesn't match the `DefaultChannelName` annotation :)

pkg/validator/am0001/default_channel.go
Ankit152: The registry implements the channels in the same way I have Implemented.
Ref: https://github.com/operator-framework/operator-registry/blob/master/pkg/registry/bundle.go#L68

pkg/validator/am0001/default_channel.go
Ankit152: Actually, It was kind of easy to do it there only. That's why I didn't created another function outside.

pkg/validator/am0001/default_channel.go
Ajpantuso: Kubernetes annotations are guaranteed to not contain whitespace (https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/#syntax-and-character-set).

pkg/validator/am0001/default_channel.go
Ajpantuso: I don't like inner functions used like this since they provide a net-neutral benefit to readability (code block is contained and semantically named, but outer function is still too big). I think the more idiomatic way to do this in Go would be to move the functionality into a `type`.

Something like the following:

```go
func channelListFromAnnotiation(ann string) channelList {
	return channelList(strings.Split(",", ann))
}

type channelList []string

func (l channelList) Contains(ch string) bool {
	for _, c := range l {
		if ch == c {
			return true
		}
	}

	return false
}

func (l channelList) String() string {
	return strings.Join(l, ", ")
}
```

Then you have have stronger cohesion in the domain of `Channels`.

Ankit152: Okay. I will do it.

garrettthomaskth: Defining a function, especially a inner function, does add some complexity. I think it might be cleaner just to write

`
defaultChannelPresent := false
for _, channel := range channels {
if defaultChannel == channel {
	defaultChannelPresent := false
        break
}
if !defaultChannelPresent {
`

garrettthomaskth: WDYT?

Ankit152: I was also thinking the same in the first place but, used inner functions instead for the same.

Ajpantuso (commented):
Code looks good however after testing against `managed-tenants` I do see that this will cause addons to fail `AM0001` which were previously successful. I would suggest running this locally against addons and see if we can address the issues in the failing bundles before committing this.

Ajpantuso:
Holding since these changes will break existing validators.

Ajpantuso (approved):
LGTM

PR #212, opened 2022-10-12
[MTSRE-725] AMO imageset and metadata api not in compliance with schema

Ankit152:

## Description

AMO imageset and metadata api not in compliance with schema.
Ref: https://github.com/mt-sre/managed-tenants-cli/blob/main/managedtenants/schemas/imageset.schema.yaml
Ref: https://github.com/mt-sre/managed-tenants-cli/blob/main/managedtenants/schemas/metadata.schema.yaml


api/v1alpha1/addonimageset_types.go
```diff
@@ -53,6 +53,14 @@ type AddonImageSetSpec struct {
 	// +optional
 	// Configs to be passed to the subscription OLM object.
 	SubscriptionConfig *mtsrev1.SubscriptionConfig `json:"subscriptionConfig"`
```
Ankit152, line 55: Shouldn't this be `config` instead of `subscriptionConfig`?
Ref: https://github.com/Ankit152/managed-tenants-cli/blob/main/managedtenants/schemas/imageset.schema.yaml#L28-L30

api/v1alpha1/addonmetadata_types.go
```diff
@@ -187,21 +187,24 @@ type AddonMetadataSpec struct {
 	SubscriptionConfig *mtsrev1.SubscriptionConfig `json:"subscriptionConfig"`
```
Ankit152, line 187: same here.

api/v1alpha1/addonimageset_types.go
Ajpantuso: Yes

Ankit152: Okay. Updating json fields.

Ajpantuso (approved):
LGTM

PR #227, opened 2022-10-19
[MTSRE-741] AM0018 Addon pullSecretName should be a part of secrets.

Ankit152:

# AM0018

## Description

Addon pullSecretName is the name of the secret under `secrets` which is supposed to be used for pulling Catalog Image under CatalogSource. As of now, there is no check to validate if pullSecretName is not nil, it is a part of secrets. 

This validator ensures the same.

### DIFF
--- a/pkg/validator/am0018/pull_secret_name.go
+++ b/pkg/validator/am0018/pull_secret_name.go
@@ -1,0 +1,63 @@
+package am0018
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/mt-sre/addon-metadata-operator/pkg/types"
+	"github.com/mt-sre/addon-metadata-operator/pkg/validator"
+)
+
+func init() {
+	validator.Register(NewPullSecretname)
+}
+
+const (
+	code = 18
+	name = "pull_secret_name"
+	desc = "Ensure that pullSecretName if not nil is present in Secrets"
+)
+
+func NewPullSecretname(deps validator.Dependencies) (validator.Validator, error) {
+	base, err := validator.NewBase(
+		code,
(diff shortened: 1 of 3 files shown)


pkg/validator/am0018/pull_secret_name_test.go
```diff
@@ -0,0 +1,102 @@
...
+				Config: &mtsrev1.Config{
+					Secrets: &[]mtsrev1.Secret{
+						{
+							Name: "test-pull-secret",
+						},
+					},
+				},
+			},
+		},
+		"pullSecretName is present and addon has multiple addons": {
```
ashishmax31, line 43:
```suggestion
		"pullSecretName is present and addon has multiple secrets defined": {
```

pkg/validator/am0018/pull_secret_name.go
```diff
@@ -0,0 +1,68 @@
...
+	for _, secret := range *secrets {
+		if pullSecretName == secret.Name {
+			present = true
+			break
+		}
+	}
+	if !present {
+		return p.Fail(fmt.Sprintf("pullSecretName %v is not present in addon secrets", pullSecretName))
+	}
+	return p.Success()
```
ashishmax31, line 67:
```suggestion
	for _, secret := range *secrets {
		if pullSecretName == secret.Name {
                         return p.Success() 
		}
	}
        return p.Fail(fmt.Sprintf("pullSecretName %v is not present in addon secrets", pullSecretName))
```

ashishmax31 (changes requested):

ashishmax31 (approved):

PR #406, opened 2023-05-18
feat: AMO API parity with mt-cli schemas

Ankit152:
## Description

Updating AMO API to keep in sync with mt-cli schemas


Ajpantuso (approved):
/lgtm

